PR #12, opened 2018-08-16
Query Service Charter

mweiden:
**Approved at the 10/11/18 DCP Architecture meeting**

This PR proposes the charter for the Query Service Module.

The Query Service is an Application Programming Interface (API) for querying Human Cell Atlas (HCA) bundle metadata with different query languages.

### DIFF
--- a/charters/QueryService/charter.md
+++ b/charters/QueryService/charter.md
@@ -45,7 +45,7 @@ kj
 * Search user interface, search application and/or analysis application work
 * Upkeep and improvement of existing indexers or ETL layers within DSS
 
-## Milestones
+## Milestones and Deliverables
 
 * **EOY 2018**
     * Architectural work done
--- a/charters/QueryService/charter.md
+++ b/charters/QueryService/charter.md
@@ -1,6 +1,4 @@
-
 # Query Service
-
 
 ## Description
 The Query Service is an Application Programming Interface (API) for querying Human Cell Atlas (HCA) bundle metadata with intuitive query syntax.
@@ -14,7 +12,7 @@ Services the Query Module provides may complement or subsume the current search 
 1. Develop a scalable public API for external indexing and other DSS event consumers
 1. Deploy all query systems as stand-alone, modular services
 1. Make DCP index and query systems agnostic to DCP schema evolution
-kj
+
 ## In-scope
 * Common index event bus and deployment infrastructure
     * Finalize support for multiple metadata indexers in DSS
--- a/charters/QueryService/charter.md
+++ b/charters/QueryService/charter.md
@@ -20,31 +20,27 @@ The Query Service is an Application Programming Interface (API) for querying Hum
     * Expose event bus for bundle lifecycle event subscriptions
     * Scalable, fault-tolerant queuing mechanism: Amazon Web Services (AWS) Simple Queuing Service (SQS) push or similar
 * Query access service API and query language
-    * Define public DSS API for modular, pluggable query engines, and hand off to DSS team for long-term ownership (this includes API to trigger index/query processing, and to handle end-user queries)
-    * In collaboration with Expression Service project, determine how Query and consumer API interact and are made complementary
-    * Explore technical methods of isolating index and query systems from metadata schema evolution
+    * Public API for modular query engine plugins
+    * In collaboration with the Expression Matrix Service, determine how Query and consumer API interact and are made complementary
 * ElasticSearch search engine
-    * Migrate ES engine to pluggable interface, and factor into separate DCP service
+    * Stand-alone ES search service, integrated into search plugin interface
     * Improve isolation from metadata schema evolution
 * Structured Query Language (SQL) query engine
+    * Stand-alone SQL query service, integrated into search plugin interface
     * SQL metadata query prototype and pilot release
         * Socialize with community, investigate use cases, and determine utility of SQL query for these use cases
         * If prototype is viable, implement on the pluggable API, and launch as separate DCP service
-        * Implementation details
-            * Database service configuration and deployment infrastructure
-            * API front-end
-            * ETL layer and database schema
-            * Event handler that adds and removes bundle metadata from the index
-            * Unit, integration, and scale tests
+    * Database service configuration and deployment infrastructure
+    * ETL layer and database schema design
+* Documentation
     * Documentation for API and query language use with metadata
     * Demonstration use of service; developer example code to enable fast and easy third party adoption of API
+* UX Research and Feedback
+    * UX validation of API and query language with tertiary analysis community
     * Provide feedback to DCP about public API suitability in building layered services (e.g. can a new Query microservice be built using only public DSS API?)
 
-### Scientific "guardrails"
-* UX validation of API and query language with tertiary analysis community
-
 ## Out-of-scope
-* Search user interface (UI), search app and/or tertiary analysis work
+* Search user interface, search app and/or tertiary analysis work
 * GraphQL or other "query" languages or environments (ES and SQL should be sufficient to prove utility of design)
 * Upkeep and improvement of existing ES indexer ETL layer
 
--- a/charters/QueryService/charter.md
+++ b/charters/QueryService/charter.md
@@ -24,7 +24,7 @@ The Query Service is an Application Programming Interface (API) for querying Hum
     * In collaboration with the Expression Matrix Service, determine how Query and consumer API interact and are made complementary
 * ElasticSearch search engine
     * Stand-alone ES search service, integrated into search plugin interface
-    * Improve isolation from metadata schema evolution
+    * Increase isolation from metadata schema evolution
 * Structured Query Language (SQL) query engine
     * Stand-alone SQL query service, integrated into search plugin interface
     * SQL metadata query prototype and pilot release
@@ -36,24 +36,24 @@ The Query Service is an Application Programming Interface (API) for querying Hum
     * Documentation for API and query language use with metadata
     * Demonstration use of service; developer example code to enable fast and easy third party adoption of API
 * UX Research and Feedback
-    * UX validation of API and query language with tertiary analysis community
+    * UX validation of API and query language with [tertiary portal](https://www.humancellatlas.org/data-sharing) community
     * Provide feedback to DCP about public API suitability in building layered services (e.g. can a new Query microservice be built using only public DSS API?)
 
 ## Out-of-scope
-* Search user interface, search app and/or tertiary analysis work
+* Search user interface, search app and/or tertiary portal work
 * GraphQL or other "query" languages or environments (ES and SQL should be sufficient to prove utility of design)
 * Upkeep and improvement of existing ES indexer ETL layer
 
 ## Milestones
-* **EOY 2018** - architectural work done, ES migrated to public API, improved resilience to metadata schema evolution.
-* **Early 2019** - pilot/preview implementation of SQL query framework; feature complete, suitable for limited use, but not scalable
-* **Mid-2019** - performant, highly scalable version, suitable for interactive data serving across entire HCA data corpus
+* **EOY 2018** - Architectural work done, pilot/preview implementation of SQL query framework; feature complete, suitable for limited use, but not scalable
+* **Early 2019** - ES migrated to public API, improved resilience to metadata schema evolution
+* **Mid-2019** - Performant, highly scalable version, suitable for interactive data serving across entire HCA data corpus
 
 
 ## Roles
 
 ### Project Lead
-[Genevieve Halliburton](mailto:ghaliburton@chanzuckerberg.com)
+[Genevieve Haliburton](mailto:ghaliburton@chanzuckerberg.com)
 
 ### Product Owner
 [Matt Weiden](mailto:mweiden@chanzuckerberg.com)
--- a/charters/QueryService/charter.md
+++ b/charters/QueryService/charter.md
@@ -3,9 +3,9 @@
 
 
 ## Description
-The Query Service is an Application Programming Interface (API) for querying Human Cell Atlas (HCA) bundle metadata with intuitive query syntax. The first iteration of the Query Service Module will investigate [ElasticSearch (ES) Query Language](https://www.elastic.co/guide/en/elasticsearch/reference/current/_introducing_the_query_language.html) and  [Structured Query Language (SQL)](https://en.wikipedia.org/wiki/SQL).
+The Query Service is an Application Programming Interface (API) for querying Human Cell Atlas (HCA) bundle metadata with intuitive query syntax.
 
-Services the Query Module provides may complement or subsume the current ES search functionality in the Data Storage Service (DSS).
+Services the Query Module provides may complement or subsume the current search functionality in the Data Storage Service (DSS).
 
 ## Objectives
 1. Build, operate, and maintain a metadata API query service that allows a variety of queries to be performed against bundle metadata stored in HCA Data Coordination Platform (DCP) DSS
@@ -14,42 +14,48 @@ Services the Query Module provides may complement or subsume the current ES sear
 1. Develop a scalable public API for external indexing and other DSS event consumers
 1. Deploy all query systems as stand-alone, modular services
 1. Make DCP index and query systems agnostic to DCP schema evolution
-
+kj
 ## In-scope
 * Common index event bus and deployment infrastructure
     * Finalize support for multiple metadata indexers in DSS
     * Expose event bus for bundle lifecycle event subscriptions
     * Scalable, fault-tolerant queuing mechanism for loading bundle lifecycle events
-* Query access service API and query language
-    * Public API for modular query engine plugins
-    * In collaboration with the Expression Matrix Service, determine how Query and consumer API interact and are made complementary
-* ElasticSearch search engine
+* Provide new iteration of or replace DSS [ElasticSearch (ES)](https://www.elastic.co) index and [query language](https://www.elastic.co/guide/en/elasticsearch/reference/current/_introducing_the_query_language.html)
     * Stand-alone ES search service, integrated into search plugin interface
     * Increase isolation from metadata schema evolution
-* Structured Query Language (SQL) query engine
+* Investigate new index technologies that provide [Structured Query Language (SQL)](https://en.wikipedia.org/wiki/SQL)
     * Stand-alone SQL query service, integrated into search plugin interface
     * SQL metadata query prototype and pilot release
         * Socialize with community, investigate use cases, and determine utility of SQL query for these use cases
         * If prototype is viable, implement on the pluggable API, and launch as separate DCP service
     * Database service configuration and deployment infrastructure
     * ETL layer and database schema design
+* Query access service API and query language
+    * Public API for modular query engine plugins
+    * In collaboration with the Expression Matrix Service, determine how Query and consumer API interact and are made complementary
 * Documentation
     * Documentation for API and query language use with metadata
     * Demonstration use of service; developer example code to enable fast and easy third party adoption of API
 * UX Research and Feedback
-    * UX validation of API and query language with [tertiary portal](https://www.humancellatlas.org/data-sharing) community
+    * UX validation of API and query language with [analysis application](https://www.humancellatlas.org/data-sharing) community
     * Provide feedback to DCP about public API suitability in building layered services (e.g. can a new Query microservice be built using only public DSS API?)
     * Provide feedback to the metadata group on schema changes that could support intuitive queries of the metadata
 
 ## Out-of-scope
-* Search user interface, search app and/or tertiary portal work
-* GraphQL or other "query" languages or environments (ES and SQL should be sufficient to prove utility of design)
-* Upkeep and improvement of existing ES indexer ETL layer
+* Search user interface, search application and/or analysis application work
+* Upkeep and improvement of existing indexers or ETL layers within DSS
 
 ## Milestones
-* **EOY 2018** - Architectural work done, pilot/preview implementation of SQL query framework; feature complete, suitable for limited use, but not scalable
-* **Early 2019** - DSS ES migrated to public API, improved resilience to metadata schema evolution
-* **Mid-2019** - Performant, highly scalable version, suitable for interactive data serving across entire HCA data corpus
+
+* **EOY 2018**
+    * Architectural work done
+    * pilot/preview implementation of SQL query framework
+    * feature complete, suitable for use in UX research, but not scalable
+* **Early 2019**
+    * DSS ES migrated to public API
+    * improved resilience to metadata schema evolution
+* **Mid-2019**
+    * Performant, highly scalable version, suitable for interactive data serving across entire HCA data corpus
 
 
 ## Roles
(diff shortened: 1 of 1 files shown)


charters/QueryService/charter.md
```diff
@@ -0,0 +1,75 @@
...
+1. Make DCP index and query systems agnostic to DCP schema evolution
+
+## In-scope
+* Common index event bus and deployment infrastructure
+    * Finalize support for multiple metadata indexers in DSS
+    * Expose event bus for bundle lifecycle event subscriptions
+    * Scalable, fault-tolerant queuing mechanism: Amazon Web Services (AWS) Simple Queuing Service (SQS) push or similar
+* Query access service API and query language
+    * Define public DSS API for modular, pluggable query engines, and hand off to DSS team for long-term ownership (this includes API to trigger index/query processing, and to handle end-user queries)
+    * In collaboration with Expression Service project, determine how Query and consumer API interact and are made complementary
```
brianraymor: _In collaboration with the Expression **Matrix** Service, determine_

charters/QueryService/charter.md
```diff
@@ -0,0 +1,75 @@
...
+    * Explore technical methods of isolating index and query systems from metadata schema evolution
```
brianraymor: It's unclear why this needs to be called out since it's part of the process to deliver in-scope products?

charters/QueryService/charter.md
```diff
@@ -0,0 +1,75 @@
...
+            * API front-end
+            * ETL layer and database schema
+            * Event handler that adds and removes bundle metadata from the index
+            * Unit, integration, and scale tests
+    * Documentation for API and query language use with metadata
+    * Demonstration use of service; developer example code to enable fast and easy third party adoption of API
+    * Provide feedback to DCP about public API suitability in building layered services (e.g. can a new Query microservice be built using only public DSS API?)
+
+### Scientific "guardrails"
+* UX validation of API and query language with tertiary analysis community
```
brianraymor: This is more of a UX or community outreach item. As proposed, the guardrails are intended to call out oversight from science governance to flag review by the Science PM.

charters/QueryService/charter.md
```diff
@@ -0,0 +1,75 @@
...
+* Common index event bus and deployment infrastructure
+    * Finalize support for multiple metadata indexers in DSS
+    * Expose event bus for bundle lifecycle event subscriptions
+    * Scalable, fault-tolerant queuing mechanism: Amazon Web Services (AWS) Simple Queuing Service (SQS) push or similar
+* Query access service API and query language
+    * Define public DSS API for modular, pluggable query engines, and hand off to DSS team for long-term ownership (this includes API to trigger index/query processing, and to handle end-user queries)
+    * In collaboration with Expression Service project, determine how Query and consumer API interact and are made complementary
+    * Explore technical methods of isolating index and query systems from metadata schema evolution
+* ElasticSearch search engine
+    * Migrate ES engine to pluggable interface, and factor into separate DCP service
```
brianraymor: This reads like an implementation detail?

charters/QueryService/charter.md
mweiden: ✅

charters/QueryService/charter.md
mweiden: deleting this

charters/QueryService/charter.md
mweiden: I will try to rephrase this to make it less about implementation

charters/QueryService/charter.md
mweiden: moving this

charters/QueryService/charter.md
```diff
@@ -0,0 +1,71 @@
...
+## Milestones
+* **EOY 2018** - architectural work done, ES migrated to public API, improved resilience to metadata schema evolution.
+* **Early 2019** - pilot/preview implementation of SQL query framework; feature complete, suitable for limited use, but not scalable
+* **Mid-2019** - performant, highly scalable version, suitable for interactive data serving across entire HCA data corpus
+
+
+## Roles
+
+### Project Lead
+[Genevieve Halliburton](mailto:ghaliburton@chanzuckerberg.com)
```
GenevieveHaliburton: Haliburton with 1 "L" :)

charters/QueryService/charter.md
```diff
@@ -0,0 +1,71 @@
...
+    * UX validation of API and query language with tertiary analysis community
+    * Provide feedback to DCP about public API suitability in building layered services (e.g. can a new Query microservice be built using only public DSS API?)
+
+## Out-of-scope
+* Search user interface, search app and/or tertiary analysis work
+* GraphQL or other "query" languages or environments (ES and SQL should be sufficient to prove utility of design)
+* Upkeep and improvement of existing ES indexer ETL layer
+
+## Milestones
+* **EOY 2018** - architectural work done, ES migrated to public API, improved resilience to metadata schema evolution.
```
GenevieveHaliburton: How firm are these milestone dates? Does first milestone imply the capacity for all DCP data to be available via public API (regardless of what data sets chosen to be made available)?

charters/QueryService/charter.md
```diff
@@ -0,0 +1,71 @@
...
+    * SQL metadata query prototype and pilot release
+        * Socialize with community, investigate use cases, and determine utility of SQL query for these use cases
+        * If prototype is viable, implement on the pluggable API, and launch as separate DCP service
+    * Database service configuration and deployment infrastructure
+    * ETL layer and database schema design
+* Documentation
+    * Documentation for API and query language use with metadata
+    * Demonstration use of service; developer example code to enable fast and easy third party adoption of API
+* UX Research and Feedback
+    * UX validation of API and query language with tertiary analysis community
```
GenevieveHaliburton: Is it clear to most readers what the line for "tertiary analysis community" is? It's unclear to me

charters/QueryService/charter.md
```diff
@@ -0,0 +1,71 @@
...
+* Common index event bus and deployment infrastructure
+    * Finalize support for multiple metadata indexers in DSS
+    * Expose event bus for bundle lifecycle event subscriptions
+    * Scalable, fault-tolerant queuing mechanism: Amazon Web Services (AWS) Simple Queuing Service (SQS) push or similar
+* Query access service API and query language
+    * Public API for modular query engine plugins
+    * In collaboration with the Expression Matrix Service, determine how Query and consumer API interact and are made complementary
+* ElasticSearch search engine
+    * Stand-alone ES search service, integrated into search plugin interface
+    * Improve isolation from metadata schema evolution
```
GenevieveHaliburton: Does "improve" here mean *increase* or *reduce* the isolation from schema evolution? As written this is unclear to me which way makes the query system more robust to future changes in metadata schema (I'd thought it was better be less isolated, but this reads like we are going to make it more isolated)

GenevieveHaliburton (approved):
Looks good! A few minor points of clarification, but overall LGTM.

charters/QueryService/charter.md
mweiden: ✅

charters/QueryService/charter.md
mweiden: I'd say not firm, given our current staffing limitations. We'll do our best and adjust the milestones if needed. I realize now that these are out of order. I am going to to push a revised order here in a sec.

charters/QueryService/charter.md
mweiden: That's what the [data-sharing](https://www.humancellatlas.org/data-sharing) document calls them. I am going to add a link.

charters/QueryService/charter.md
mweiden: More isolated is correct. I will use the word `Increase`

charters/QueryService/charter.md
```diff
@@ -0,0 +1,71 @@
+
+# Query Service
+
+
+## Description
+The Query Service is an Application Programming Interface (API) for querying Human Cell Atlas (HCA) bundle metadata with different query languages.
+
+## Objectives
+1. Build, operate, and maintain a metadata API query service that allows a variety of queries to be performed against bundle metadata stored in HCA Data Coordination Platform (DCP) Data Storage Service (DSS)
+1. Develop a query language specification and metadata Extract, Transform, Load (ETL) pattern to accommodate intuitive querying
```
tburdett: This seems like two objectives in one, and the ETL pattern is somewhat an implementation detail. The metadata charter (currently proposed) specifically calls out that extending and improving the schema based on feedback (of which intuitive querying would presumably encompass) is in scope. I can see a risk here that query usecases are never propagated back to the metadata team

charters/QueryService/charter.md
```diff
@@ -0,0 +1,71 @@
...
+1. Conduct user experience (UX) research to validate query language usability, understand remaining usability gaps in DCP data searchability and discoverability
+1. Develop a scalable public API for external indexing and other DSS event consumers
+1. Migrate existing ElasticSearch (ES) query subsystem to this public API
+1. Deploy all query systems as stand-alone, modular services
+1. Make DCP index and query systems agnostic to DCP schema evolution
```
tburdett, line 14: It's unclear to me how you will be able to deliver on this objective. I see two broad types of metadata changes; syntactic type changes (renaming or moving of fields) and semantic changes (splitting, merging, addition of new required fields etc). These are described in much more detail in the [metadata-schema](https://github.com/HumanCellAtlas/metadata-schema/blob/master/docs/evolution.md) repo. I can see possibilities of coding out of syntactic changes (and the metadata team have proposed several options to enable schema introspection to avoid a "lowest common denominator" approach) but I do not see how it is possible to be agnostic to semantic schema evolution; furthermore, these are likely to be the interesting changes driven by evolving scientific requirements

charters/QueryService/charter.md
```diff
@@ -0,0 +1,71 @@
+
+# Query Service
+
+
+## Description
+The Query Service is an Application Programming Interface (API) for querying Human Cell Atlas (HCA) bundle metadata with different query languages.
```
tburdett: It would be helpful for me here if you clarify here the relation to the datastore, and specifically the consumer API as illustrated on the overview diagram? Is the query service intended to be a subcomponent of the DSS consumer API that pertains only to metadata, or a service built on top of it?

charters/QueryService/charter.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Objectives
+1. Build, operate, and maintain a metadata API query service that allows a variety of queries to be performed against bundle metadata stored in HCA Data Coordination Platform (DCP) Data Storage Service (DSS)
+1. Develop a query language specification and metadata Extract, Transform, Load (ETL) pattern to accommodate intuitive querying
+1. Conduct user experience (UX) research to validate query language usability, understand remaining usability gaps in DCP data searchability and discoverability
+1. Develop a scalable public API for external indexing and other DSS event consumers
+1. Migrate existing ElasticSearch (ES) query subsystem to this public API
```
tburdett: This feels more like a milestone than a charter objective

charters/QueryService/charter.md
```diff
@@ -0,0 +1,71 @@
...
+1. Deploy all query systems as stand-alone, modular services
+1. Make DCP index and query systems agnostic to DCP schema evolution
+
+## In-scope
+* Common index event bus and deployment infrastructure
+    * Finalize support for multiple metadata indexers in DSS
+    * Expose event bus for bundle lifecycle event subscriptions
+    * Scalable, fault-tolerant queuing mechanism: Amazon Web Services (AWS) Simple Queuing Service (SQS) push or similar
```
tburdett: This first statement seems heavily implementation specific; why is this being called out? I couldn't figure out why I need this information

charters/QueryService/charter.md
mweiden: ✅:
1) Removing the ETL as a goal assuming it is an implementation detail
2) Adding the metadata group to the feedback section

charters/QueryService/charter.md
mweiden: I think your concerns are on target. We will not be able to isolate _everyone_ from changes in the metadata. Clients will still need to deal with changes in the metadata. However, we have some ideas on how to make the _index and query systems_ robust to these changes.

charters/QueryService/charter.md
mweiden: You're right. In fact, it is already in the milestones. Removing this.

charters/QueryService/charter.md
mweiden: ✅

charters/QueryService/charter.md
mweiden: ✅

mweiden:
@tburdett @brianraymor @GenevieveHaliburton Please see revisions. I believe that they cover all of your comments.

charters/QueryService/charter.md
```diff
@@ -0,0 +1,73 @@
...
+    * SQL metadata query prototype and pilot release
+        * Socialize with community, investigate use cases, and determine utility of SQL query for these use cases
+        * If prototype is viable, implement on the pluggable API, and launch as separate DCP service
+    * Database service configuration and deployment infrastructure
+    * ETL layer and database schema design
+* Documentation
+    * Documentation for API and query language use with metadata
+    * Demonstration use of service; developer example code to enable fast and easy third party adoption of API
+* UX Research and Feedback
+    * UX validation of API and query language with [tertiary portal](https://www.humancellatlas.org/data-sharing) community
```
brianraymor: Regarding the use of _tertiary_ - please see Jodi's note - https://humancellatlas.slack.com/archives/C53S0GBS7/p1534526020000100

charters/QueryService/charter.md
kislyuk: @tburdett appreciate your skepticism here, and it's totally warranted. The intent of our approach here is to build a *general purpose* metadata indexing layer with zero-configuration accommodation of schema changes. This means it should be able to deal with schema heterogeneity without having to reprogram the underlying database or API with custom schemas or migrations whenever upstream metadata schema changes. That does not mean the downstream metadata consumers stop caring about metadata changes; both syntactic and semantic changes will still pass through to them and require updates (separately, the way to deal with those is by versioning the schema generations and filtering metadata by major schema versions).

kislyuk (approved):

charters/QueryService/charter.md
```diff
@@ -0,0 +1,73 @@
...
+## Description
+The Query Service is an Application Programming Interface (API) for querying Human Cell Atlas (HCA) bundle metadata with intuitive query syntax. The first iteration of the Query Service Module will investigate [ElasticSearch (ES) Query Language](https://www.elastic.co/guide/en/elasticsearch/reference/current/_introducing_the_query_language.html) and  [Structured Query Language (SQL)](https://en.wikipedia.org/wiki/SQL).
+
+Services the Query Module provides may complement or subsume the current ES search functionality in the Data Storage Service (DSS).
+
+## Objectives
+1. Build, operate, and maintain a metadata API query service that allows a variety of queries to be performed against bundle metadata stored in HCA Data Coordination Platform (DCP) DSS
+1. Develop a query language specification for intuitive querying of HCA metadata
+1. Conduct user experience (UX) research to validate query language usability, understand remaining usability gaps in DCP data searchability and discoverability
+1. Develop a scalable public API for external indexing and other DSS event consumers
```
kozbo, line 12: Does the generic event system provided in the Data Store Charter provide this? "Transition Data Store Subscriptions/Eventing services from the current dependence on Elastic Search Percolate to the AWS and GCP cloud infrastructure."

mweiden: That's a good question. I've discussed this with @kislyuk and think there is a way forward on this. Let's discuss during the Arch meeting Monday.

charters/QueryService/charter.md
tburdett: Great! 100% agreed zero config metadata indexing is the goal. I think we're on the same page, but there's probably a wider conversation here that should involve those on the query service/indexing side and the metadata team.

I think at this point I'd like to call out the prospect of automatic migration of older data (using older schemas) to the most recent schema wherever possible. Is this something you guys would envisage as being in scope for the query service, the DSS, the data portal? I don't feel ingest is the correct home for this; it de facto requires retention of old data which is out of scope for ingest

charters/QueryService/charter.md
```diff
@@ -0,0 +1,73 @@
+
+# Query Service
+
+
+## Description
+The Query Service is an Application Programming Interface (API) for querying Human Cell Atlas (HCA) bundle metadata with intuitive query syntax. The first iteration of the Query Service Module will investigate [ElasticSearch (ES) Query Language](https://www.elastic.co/guide/en/elasticsearch/reference/current/_introducing_the_query_language.html) and  [Structured Query Language (SQL)](https://en.wikipedia.org/wiki/SQL).
```
TimothyTickle: I think the second sentence here (about ES and SQL) is an objective or milestone but not a description.

mweiden: Good point. Moving this.

charters/QueryService/charter.md
```diff
@@ -0,0 +1,79 @@
...
+    * Database service configuration and deployment infrastructure
+    * ETL layer and database schema design
+* Query access service API and query language
+    * Public API for modular query engine plugins
+    * In collaboration with the Expression Matrix Service, determine how Query and consumer API interact and are made complementary
+* Documentation
+    * Documentation for API and query language use with metadata
+    * Demonstration use of service; developer example code to enable fast and easy third party adoption of API
+* UX Research and Feedback
+    * UX validation of API and query language with [analysis application](https://www.humancellatlas.org/data-sharing) community
```
brianraymor: Is the intention to reference the HCA Analysis Working Group which does not have a handy link? Otherwise, this is basically a reference to the DCP community?

mweiden: No, not the Analysis Working Group. Somewhere, I think Jodi and MattS suggested that we use the term _analysis portals_ to refer to "Red Boxes."

brianraymor: Doh. This terminology change - https://humancellatlas.slack.com/archives/C53S0GBS7/p1534526020000100 -Unfortunately,  I don't see how the HCA reference in its current state clarifies that term at all.

mweiden: OK, I will remove it

charters/QueryService/charter.md
```diff
@@ -0,0 +1,79 @@
...
+Services the Query Module provides may complement or subsume the current search functionality in the Data Storage Service (DSS).
+
+## Objectives
+1. Build, operate, and maintain a metadata API query service that allows a variety of queries to be performed against bundle metadata stored in HCA Data Coordination Platform (DCP) DSS
+1. Develop a query language specification for intuitive querying of HCA metadata
+1. Conduct user experience (UX) research to validate query language usability, understand remaining usability gaps in DCP data searchability and discoverability
+1. Develop a scalable public API for external indexing and other DSS event consumers
+1. Deploy all query systems as stand-alone, modular services
+1. Make DCP index and query systems agnostic to DCP schema evolution
+kj
```
kislyuk: Looks like a typo here

charters/QueryService/charter.md
```diff
@@ -0,0 +1,79 @@
+
```
kislyuk: Please remove whitespace before the first heading.

charters/QueryService/charter.md
mweiden: Good catch. That's my `vim` tick.

charters/QueryService/charter.md
```diff
@@ -0,0 +1,77 @@
+# Query Service
+
+## Description
+The Query Service is an Application Programming Interface (API) for querying Human Cell Atlas (HCA) bundle metadata with intuitive query syntax.
+
+Services the Query Module provides may complement or subsume the current search functionality in the Data Storage Service (DSS).
+
+## Objectives
+1. Build, operate, and maintain a metadata API query service that allows a variety of queries to be performed against bundle metadata stored in HCA Data Coordination Platform (DCP) DSS
+1. Develop a query language specification for intuitive querying of HCA metadata
```
kbergin: Are these all '1's on purpose? i.e. is it indicating priority or is it supposed to be a numerical list?

mweiden: @kbergin this is markdown convention. It renders as an ordered list.

kbergin: Ooh, today I learned! Thanks!

charters/QueryService/charter.md
```diff
@@ -0,0 +1,79 @@
...
```
diekhans: Define the target users of intuitive would bound this objective.  I would assume it is programmers, not end-user.

tburdett (approved):
This looks fine now. There's a couple of final comments I have that you can take or leave as you like... 😃 

- It's still kind of unclear to me where the line between bundles in the DSS ends and queries in the query service begins. My assumption was that bundles exist as a way to deliver "pre-canned" query responses to users by anticipating the granularity at which they need the data. It makes sense to me that there will be queries that don't map 1:1 onto existing bundle structures, but how much responsibility do the query service and the data store take for improving this? Specifically - if 99% of user requests can be shown to require the query service to slice and dice the data differently to existing bundles, why not just change the bundle structure (or add new bundle structures)? The Data Store Charter specifically allows for this and the Query Service will have valuable query data that could be mined to provide evidence of user needs
- I like the additional focus on defining the users of the Query Service. You could almost go further and include this in the description right at the top. Is it intended, for example, that advanced users could use a power interface build a query in the Data Portal that directly queries the Query Service?
- This charter has a more technical focus than other charters and some of the specifics could be left out
- I'm not sure SQL is necessarily the right choice of language here given the variable nature of metadata graphs. If I was implementing this, I'd explore graph query languages - you could look at GraphQL, OpenCypher or even SPARQL.

charters/QueryService/charter.md
```diff
@@ -0,0 +1,79 @@
...
+* Provide new iteration of or replace DSS [ElasticSearch (ES)](https://www.elastic.co) index and [query language](https://www.elastic.co/guide/en/elasticsearch/reference/current/_introducing_the_query_language.html)
+    * Stand-alone ES search service, integrated into search plugin interface
+    * Increase isolation from metadata schema evolution
+* Investigate new index technologies that provide [Structured Query Language (SQL)](https://en.wikipedia.org/wiki/SQL)
+    * Stand-alone SQL query service, integrated into search plugin interface
+    * SQL metadata query prototype and pilot release
+        * Socialize with community, investigate use cases, and determine utility of SQL query for these use cases
+        * If prototype is viable, implement on the pluggable API, and launch as separate DCP service
+    * Database service configuration and deployment infrastructure
+    * ETL layer and database schema design
```
diekhans, line 38: Do specific implementations (SQL, ElasticSearch) actually belong at in the charter?   I would agree with Tony that some of the modern query languages might be worth investigation.

mweiden:
@tburdett @diekhans Good thoughts. I've added a users section. Further, I don't think that the charter precludes spikes to investigate other areas/technologies to make sure we're on the right track. I've made slight wording changes to reflect this.

mweiden:
This was approved in Arch this morning. Merging now.

PR #14, opened 2018-08-17
DCP Communications Team Charter

matthewspeir:
Charter for DCP Communications Team. Ready for review. @jodihir @brianraymor

### DIFF
--- a/charters/CommunicationsTeam/charter.md
+++ b/charters/CommunicationsTeam/charter.md
@@ -3,22 +3,13 @@
 
 
 ## Description
-The HCA DCP Communications team coordinates and publishes public communications through
-various channels for the DCP as a whole. We work with the different groups within the DCP
-to ensure that each group is properly represented in external communications. We work with
-the HCA communications representatives if/when needed to ensure that DCP communications
-are consistent with and support those of the HCA. We monitor the DCP Data Portal
-documentation and update as needed. We develop training materials and events to teach
-people about the HCA DCP and how to use it.
+The Human Cell Atlas (HCA) Data Coordination Platform (DCP) Communications team coordinates and publishes public communications through various channels for the DCP as a whole. We work with the different groups within the DCP to ensure that each group is properly represented in external communications. We work with the HCA communications representatives if/when needed to ensure that DCP communications are consistent with and support those of the HCA. We monitor the DCP Data Portal documentation and update as needed. We develop training materials and events to teach people about the HCA DCP and how to use it.
 
 ## Objective
 * Provide clear and accurate information to the public regarding DCP activities
 * Provide educational materials to help people to take advantage of DCP tools and HCA data
-* Function as the front line for communications between the DCP and legal or media entities,
-triaging requests or issues to the appropriate DCP or HCA leadership and possibly creating
-first drafts for their review (eg: press releases).
-* Support users and ensure that questions regarding DCP tools or HCA data are forwarded
-onto the appropriate group
+* Function as the front line for communications between the DCP and legal or media entities, triaging requests or issues to the appropriate DCP or HCA leadership and possibly creating first drafts for their review (eg: press releases).
+* Support users and ensure that questions regarding DCP tools or HCA data are forwarded onto the appropriate group
 
 ## In-scope
 * Update and maintain various public communication channels, for example:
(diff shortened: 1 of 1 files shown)


charters/CommunicationsTeam/charter.md
```diff
@@ -0,0 +1,64 @@
...
+## In-scope
+* Update and maintain various public communication channels, for example:
+    * Social media channels
+    * Direct user communication
+    * Support applications (eg, Zendesk)
+* Develop and host educational materials  and events describing the DCP components
+* Monitor the state of our communications to ensure they are up-to-date and complete
+* Develop long-term plans for HCA DCP communication and estimate the resources needed to execute these plans
+
+### Scientific "guardrails"
```
brianraymor, line 23: should be -> must be

matthewspeir: Good catch. Thanks! I've updated the doc and committed the change.

charters/CommunicationsTeam/charter.md
```diff
@@ -0,0 +1,64 @@
...
+If we are communicating about any of the scientific activities of the HCA, the communication should be coordinated with the HCA project’s communications team.
+
+## Out-of-scope
+* Any communications for the HCA related to the scientific arm of the HCA project. 
+* UX research that informs us about the success of the portal or desired portal engineering 
+
+## Milestones
+* Announcement mailing list assembled
```
jonahcool, line 31: Is there a strong rationale for why we would want an independent DCP mailing list? It would be convenient but I also worry that it could become confusing or fragmented.

charters/CommunicationsTeam/charter.md
```diff
@@ -0,0 +1,64 @@
...
+    * Pull email list form HCA registry of interest and allow people to opt-in to future emails
```
jonahcool: "Pull email list form HCA registry...." to "Pull email list from HCA registry.." Also, see point above about risking fragmentation - would be great to get the registry/dcp lists in sync. Totally understand if there are major logistical hurdles to this but think we should clarify.

charters/CommunicationsTeam/charter.md
matthewspeir: Hmm, excellent question. 

I was imagining that we would announce things like DCP feature release, API updates, etc. to a DCP-specific mailing list. Maybe @jodihir has other ideas about what types of things would go to a DCP-specific mailing list. Do you think that these types of things would relevant to a more general HCA mailing list?

jonahcool (commented):
I found this charter very clear and helpful - it touched on critical areas and raised a few interesting questions. A few comments throughout identify one primary concern, which is that the DCP should avoid duplicating effort or instilling confusion by creating similar but distinct communications resources from the HCA. It will be critical for DCP communications to capture and share information, as outlined, with the community - just think that leveraging existing resources and building them collectively is important.

charters/CommunicationsTeam/charter.md
jonahcool: Great point - perhaps a subset does indeed make sense. Suppose my preference would be to try and generate this in parallel with the HCA email list (e.g. by someone opting in via the main registry) as opposed to forking and creating a new one.

TimothyTickle: I think it makes a lot of sense to leverage a common onboarding for people to communication channels but I believe coordinated but separate communication efforts are important. We will be communicating different kinds of information at a different velocity than the HCA science arm. It is important people can opt in an out of the different communications and that our communications team is enabled to proceed as they need and do not have barriers when communicating DCP related material.

jodihir: I agree and I like Jonah's idea of generating the DCP email list via an option for users from the main registry. In addition to the ease for users, it gives us an opportunity to clarify to users what the DCP does as part of the HCA.

jodihir:
Thanks @jonahcool for your thoughts on this.  We will add with some wording to the charter that speaks directly to your point about not duplicating effort with HCA communications; making sure what we are doing is distinct and yet collaborative.

charters/CommunicationsTeam/charter.md
```diff
@@ -0,0 +1,64 @@
+
+# Data Coordination Platform Communications Team
+
+
+## Description
+The Human Cell Atlas (HCA) Data Coordination Platform (DCP) Communications team coordinates and publishes public communications through various channels for the DCP as a whole. We work with the different groups within the DCP to ensure that each group is properly represented in external communications. We work with the HCA communications representatives if/when needed to ensure that DCP communications are consistent with and support those of the HCA. We monitor the DCP Data Portal documentation and update as needed. We develop training materials and events to teach people about the HCA DCP and how to use it.
```
jodihir, line 6: Suggested change to the sentence that starts "We work with the HCA communications representatives as needed.."  to the following:
We work with the HCA communications representatives as needed to ensure that DCP communications are distinct from but consistent with those of the HCA, and that we are sharing resources when possible and appropriate to avoid duplication of efforts and user confusion."

matthewspeir: I think that sounds great. I think we can go ahead with moving this to 'charter-final-review' as I believe we've addressed all of the comments.

brianraymor:
@jodihir and @matthewspeir - is this charter withdrawn until the broader proposal is funded?

matthewspeir:
@brianraymor Yes, I think we can withdraw this for now. Should I delete this pull request? (Or I guess, how do I delete this pull request?)

brianraymor:
Annotating as _withdrawn_ and closing.

PR #18, opened 2018-08-21
Added basic guidance to charter template sections

### LINKED ISSUE #15: Objective or Objectives
I think in the context of  our charters Objective as found in https://github.com/HumanCellAtlas/dcp-community/blob/master/charters/charter-template.md tends to be used in the plural rather than singular

Can this be updated to the plural Objectives?

brianraymor:
Closes #15

* Also renames **Milestones** to **Milestones and Deliverables**

### DIFF
--- a/charters/charter-template.md
+++ b/charters/charter-template.md
@@ -4,9 +4,15 @@
 
 ## Description
 
-## Objective
+_Describe the vision of the project in **2-3** sentences. Consider this section as your "elevator pitch" to new community members._
+
+## Objectives
+
+_Describe the incremental, observable, and measurable high-level outcomes of the project_
 
 ## In-scope
+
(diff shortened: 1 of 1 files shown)


kozbo:
I like the most recent additions. One comment I have is that it is hard to know what to include/exclude. Should a charter be just about feature delivery? Should we also include tasks that are an on-going responsibility? Those types of tasks never complete, thus cannot be marked as done.

**Brian responds:**

@bkmartinjr and I had a related conversation yesterday about _operational roles_. Bruce can comment further, but we were thinking:

* Document operational roles in **Objectives** and **In-Scope**
* Assume that you're doing devops as long as you're in active development
* When the development milestones end, then re-charter
* When there are no further development milestones, then pivot to pure ops

Are there other _never complete_ cases that you had in mind?

PR #31, opened 2018-09-25
Unity Charter

TimothyTickle:


TimothyTickle:
Note, will change charter-template.md to charter.md on the next set of edits. Realized after merging the fork but figured I could hear other edits to do at the same time.

PR #34, opened 2018-09-25
Metadata Integration Testing

mweiden:
For more information, please read the RFC text.

Suggest: 2 week review timebox (end of timebox 2018-10-19)


hewgreen (dismissed):
(I meant to comment here not approve) From my perspective on the metadata team, this is really nice to see. I'd like to see some more details about dcp integration tests vs a metadata test suite. I'm not sure if your doc fully separates these?

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,106 @@
...
+
+* "Test then merge"
+* PR status will not require wranglers to know about downstream systems
+* Small non-breaking changes can be merged quickly and with confidence
+* Greatly reduces probability of deployment stages past `dev` breaking
+
+#### Cons
+
+* May introduce another deployment stage
+* Treats metadata schema integration differently from other DCP projects
```
malloryfreeberg, line 100: Why is it a Con that the metadata schemas are treated differently than other DCP components?

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,106 @@
...
+
+* May be difficult to revert test versions of components and metadata (not all infra deployments are easily reversible)
+* Requires "serializing" the dcp-wide integration test so that one PR can run after another
+
+### Design Alternative #2 
+
+Software development proceeds in the dev branch of each component independently.
+When a metadata schema version has passed unit tests in `dev` and its committers wish to promote it, they make a PR against their integration branch.
+The PR triggers a dcp-wide test in the integration environment.
+Metadata schema changes can be merged from `dev` to `integration` only if the integration test suite passes.
```
malloryfreeberg, line 84: > When a metadata schema version has passed unit tests in `dev` and its committers wish to promote it, they make a PR against their integration branch. The PR triggers a dcp-wide test in the integration environment. Metadata schema changes can be merged from `dev` to `integration` only if the integration test suite passes.

> Data with the new schema version should not be uploaded into the `integration` or subsequent deployment stages until the test passes and the `master` branch is promoted to `integration`.

These two statements seems like they have the potential to massively delay data contributors being able to submit data to the DCP. For many reasons, this delay is not ideal. What is that rationale behind requiring all components of the DCP pass integration with the new schema before a data contributor can submit data? I'm confused as to why a contributor could not submit a dataset requiring a new field if - as an extreme example - the Data Browser fails to incorporate the new field as a facet (which could cause the integration test to fail). Any thoughts on this?

rfcs/text/0000-metadata-integration-test.md
mweiden: It's not necessarily a con. I don't see it as such. It just felt important to acknowledge the design as asymmetrical.

rfcs/text/0000-metadata-integration-test.md
mweiden: Yes!

Unfortunately, there is some context missing here. We have been working with @tburdett on a draft development plan to decouple ingest from new schema support in downstream applications. Please DM me for the details. 

This brings me to the rationale... As outlined in the **Motivation** section, the rationale for these changes is to keep the development environment stable while we iterate towards having decoupling ingest/upload/storage from schema support in downstream apps. The system does _not_ currently support this. In short, creation of DSS events is locked to bundle creation and there is no way to revisit events that fail to be processed downstream if there is a breaking change in the metadata schema.

Having a blocking integration test is a temporary fix until we can engineer around this.

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Detailed Design
+
+Below we describe a design that will transition DCP from the current "waterfall" software/metadata integration model to a more observable and interactive integration model, informed by integration tests that display the true compatibility state of the system.
+
+Setting this plan will also better define the interaction between the metadata team and the rest of the DCP. Processes and responsibilities for resolving any integration issues that arise will be clear.
+
+### Design Alternative #1
+
+Software development proceeds in the dev branch of each component independently. Changes are merged into integration branch of each component only through PRs which cause system-wide integration tests to run. PRs are merged automatically if tests pass (but can be reverted if downstream breakage occurs). PRs may not be merged if tests do not pass.
```
brianraymor: Can you clarify why _PRs may not be merged if tests do not pass_ is a MAY NOT rather than a MUST NOT? Under what conditions would be the PRs be merged? Not sure if it's related to the **Cons** - _May be difficult to revert ..._ case.

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,106 @@
...
+### Design Alternative #2 
+
+Software development proceeds in the dev branch of each component independently.
+When a metadata schema version has passed unit tests in `dev` and its committers wish to promote it, they make a PR against their integration branch.
+The PR triggers a dcp-wide test in the integration environment.
+Metadata schema changes can be merged from `dev` to `integration` only if the integration test suite passes.
+
+![Option 2](../images/0000-metadata-integration-test-opt2.png)
+
+Data with the new schema version should not be uploaded into the `integration` or subsequent deployment stages until the test passes and the `master` branch is promoted to `integration`.
```
brianraymor, line 88: _**should not** be uploaded_ or **must not**? See [RFC2119](https://www.ietf.org/rfc/rfc2119.txt):

>4. SHOULD NOT   This phrase, or the phrase "NOT RECOMMENDED" mean that
   there may exist valid reasons in particular circumstances when the
   particular behavior is acceptable or even useful, but the full
   implications should be understood and the case carefully weighed
   before implementing any behavior described with this label.

tburdett (changes requested):
I have a number of concerns over this RFC. @malloryfreeberg's [comment](https://github.com/HumanCellAtlas/dcp-community/pull/34#discussion_r223634407) summarises well my main objection. It also doesn't make sense to me to enshrine a temporary workaround into the DCP design via RFC.

Given current priorities around getting data through the system for the beta and how busy everyone is currently with this task, I don't think the October 19th timebox for review makes much sense. I'd like to propose tabling this RFC until after the face to face, and after the main design decision about how to decouple DCP clients from newer metadata is established.

rfcs/text/0000-metadata-integration-test.md
mweiden: I think this one can be a "should not."

mweiden:
@tburdett @malloryfreeberg Sure. We can table this until the F2F.

We need to keep planning for solving this problem so that it doesn't come back to bite us, though. I will follow up on this with you.

rfcs/text/0000-metadata-integration-test.md
brianraymor: If that is the case, then the _valid reasons in particular circumstances_ could be documented; otherwise, there's a tendency to interpret *SHOULD NOT* as *OPTIONAL*.

mweiden:
After our discussion that the F2F, there are some edits that need to be made here and some additional context to add. In order to keep `dcp-comminity` PRs clean and to focus the discussion on the new proposal, I am going to close this PR and resubmit the revision when it's ready.

PR #54, opened 2018-12-16
Rfc metadata integration test

diekhans:
Added a section on using version numbers to allow data to proceed part way through the pipeline.

### DIFF
--- a/rfcs/text/0000-metadata-integration-test.md
+++ b/rfcs/text/0000-metadata-integration-test.md
@@ -50,7 +50,7 @@ Data with the new schema version should not be uploaded into the `integration` o
 * Treats metadata schema integration differently from other DCP projects
 * Requires test data to that use the new schema features
 
-## Alternative design
+## Automatic merge alternative
 
 Software development proceeds in the dev branch of each component independently. Changes are merged into integration branch of each component only through PRs which cause system-wide integration tests to run. PRs are merged automatically if tests pass (but can be reverted if downstream breakage occurs). PRs must not be merged if tests do not pass.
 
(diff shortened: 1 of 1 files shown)


rfcs/text/0000-metadata-integration-test.md
```diff
@@ -73,6 +74,26 @@ Data with the new schema version should not be uploaded into the `integration` o
...
+
+#### Pros:
+* Easy to implement, especially for metadata developers.
+* Good fit for the serial nature of the pipeline.
+* Works very will for simple changes that can be done quickly
+
+#### Cons:
+* Require serial integration of metadata changes in version number order, requiring quick response to incompatibilities.
+* Complex, time consuming changes could stall unrelated metadata updates from being published.
+
```
mweiden, line 98: I would add another (big) con here. Data that is filtered out will often have to be replayed manually through the system or from component to component.

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -48,6 +48,7 @@ Data with the new schema version should not be uploaded into the `integration` o
 
 * May introduce another deployment stage
 * Treats metadata schema integration differently from other DCP projects
+* Requires test data to that use the new schema features
 
 ## Alternative design
```
mweiden, line 54: Please add a title to the first alternative design to differentiate the two. Perhaps `### All projects must PR before merging into integration environment`

mweiden (changes requested):
Just two small tweaks. Otherwise LGTM 👍

rfcs/text/0000-metadata-integration-test.md
diekhans: My intention here is that data is just queued waiting for the next stage to support it.  Once supported (a commit of an updated version list), then the queued submission can be processed.  I will updated the text to reflect this

PR #57, opened 2019-01-11
RFC Deletion Process

Bento007:
~Last call Feb. 22, 2019~
**March 15**: Last call for oversight review


brianraymor:
@Bento007 - is this RFC _in progress_ and not ready for community review?

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,279 @@
+### DCP PR:
+
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC number and this pull request in the format:*
+
+`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+
+# RFC Deletion and deprecation of data in the DCP
```
xbrianh: Depreciation doesn't appear to be addressed in this RFC.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,279 @@
...
+Reasons for deletion of (meta)data are
+
+* consent to share (meta)data derived from a donor's specimens is withdrawn by
+  that donor after the (meta)data was submitted to DCP
+  
+* that consent was never given but this fact is revealed only after the
+  (meta)data was submitted to the DCP
+  
+* storing or processing the (meta)data causes a serious disruption of service
+  within the DCP
```
xbrianh: Valid data or metadata that causes our services to fail indicates a serious design flaw or bug. We should not need to resort to deletion in order to fix our system.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,276 @@
...
+     "details": "<additional info>",
+     "admin_email": "<admin>"
+   }
+   ```
+   The tombstone causes `GET /bundle/{uuid}` to return a 404. If `{version}` 
+   signifies the latest version of the bundle
+   identified by `{uuid}`, requests for `GET /bundle/{uuid}` without a version
+   also return a 404. The same applies to HEAD requests, both versioned and
+   unversioned. The file versions referenced by the deleted bundle version are
+   not immediately affected by the deletion and remain accessible. The bundle and associated files are added to
```
xbrianh: The deletion daemon is an implementation detail of our deletion policies. I don't think we need to explicitly reference it in the RFC.

rfcs/text/0000-dss-deletion-daemon.md
Bento007: will remove

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,276 @@
...
+   signifies the latest version of the bundle
+   identified by `{uuid}`, requests for `GET /bundle/{uuid}` without a version
+   also return a 404. The same applies to HEAD requests, both versioned and
+   unversioned. The file versions referenced by the deleted bundle version are
+   not immediately affected by the deletion and remain accessible. The bundle and associated files are added to 
+   the deletion daemons queue. The deletion queue is an AWS SQS.
+   
+   A notification is sent out to subscribers of datastore deletions when a bundle is tombstoned.
+
+4) The deletion daemon is an AWS Lambda with privilages to deleted files from the replicas. An administrator
```
xbrianh: This section covers **physical deletes**?

I don't feel we should reference a deletion daemon, deletion markers, bucket lifecycles, etc. We should discuss policies/interfaces while leaving ourselves free to architect their implementation.

The policies appear to be:
- When will data be made inaccessible (hopefully immediately)
- Grace periods/reversals
- logging

rfcs/text/0000-dss-deletion-daemon.md
Bento007: but the flaw may not be from the DSS and it might easier to delete the files to avoid impeding the progress of others. Deletion in general should never happen. These are worst case scenarios.

rfcs/text/0000-dss-deletion-daemon.md
Bento007: fixed

rfcs/text/0000-dss-deletion-daemon.md
Bento007: fixed

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+## Motivation
+
+*Describe the user or technical need in detail [with alignment to the DCP roadmap priorities where possible]. Link prior community discussions to demonstrate support for this RFC.*
+
+One of the DCP's primary objectives is to ensure the durability of the
+(meta)data it contains and to maintain a complete record of the changes made to
+it. The DCP achieves this by never overwriting a (meta)data file but instead
+treating an update to such a file as the addition of a new version of that
+file. And instead of physically removing a file version, the DCP hides that
+file version behind a deletion marker.
```
kozbo: I don't think that an update always result in the old version being tagged with a deletion marker. I think that deletion of old versions is more the exception than the rule. This text reads like it is always the case.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+
+Under certain circumstances however, the *physical deletion* of a file version
+is required. The first section of this specification describes the
+circumstances and mechanisms for logical and physical deletion of files and the
+bundles that contain them.
+
+### User Stories
+
+*Share the [User Stories](https://www.mountaingoatsoftware.com/agile/user-stories) motivating this RFC.*
+As an operator of the DSS, I found unconsented data in the DSS that must be removed, to protect user's privacy.
```
kozbo: should we add a clause here that we will want to remove unconsented data and its derivations like secondary analysis data?

kozbo: Ah, I now see the "Derived data" section below. Not sure if we want to mention it up here too, your choice.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+versions. Administrators are responsible for tracking down data bundles with
+data derived from data in deleted bundles and to issue deletion requests for
+them.
+
+### Tiered storage
+
+Physical deletion of blobs is effective across all tiered storage classes in
+the underlying blob store. Additional cost may be incurred for deleting data recently moved to a
+slower access storage tier. Deletion occur There is no additional delays cause when deleting from a 
+tiered storage system such as AWS glacier.
```
kozbo: some grammar issues with this sentence

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+
+### Undoing logical deletions
+
+A bundle can be restored using `restore_bundle.py <uuid.<version>` if has not been physically deleted. This process
+uses the info from the deletion table to removes tombstones, and prevent permanent deletion. Subscribers of new
```
kozbo: "if it has not been..."

kozbo: "from the deletion table to remove tombstones,"

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+TODO: enable bucket versioning and lifecycle rules for replicas.
+
+TODO: Remove `Bundle/Delete` API from DSS.
+
+### Acceptance Criteria [optional]
+
+*Acceptance criteria are the conditions that a RFC must satisfy to be accepted by users or other stakeholders.*
+
+* When a bundle is tombstoned it is made immediately unavailable to users using search.
+* Maintainers of external indices shall removed deleted data from its index when a bundle tombstone notification is received.
```
kozbo: not sure on the grammar here, suggest "remove deleted data from their indices"

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+### Unresolved Questions
+
+- *What aspects of the design do you expect to clarify further through the RFC review process?*
+- *What aspects of the design do you expect to clarify later during iterative development of this RFC?*
+
+### Drawbacks and Limitations [optional]
+
+* This introduces the ability to permanently destroy data from the DSS. 
+* There is no limit on the amount of data that can be rapidly permanently deleted.
+* Permanently deletion does not occur until after the grace period has elapsed. This could
```
kozbo: "Permanent deletion"

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+  be problematic if files need to be removed sooner. 
+* If two bundles share files and only one bundle is deleted, this causes the remaining bundle to be incomplete.
```
kozbo: Is this the same as the "Dangling bundle" example above?

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+* Bundles are not completely removed from the DSS until after the grace period has elapsed.
+
+### Unresolved Questions
+
+- *What aspects of the design do you expect to clarify further through the RFC review process?*
+- *What aspects of the design do you expect to clarify later during iterative development of this RFC?*
+
+### Drawbacks and Limitations [optional]
+
+* This introduces the ability to permanently destroy data from the DSS.
```
justincc: IMHO this is unavoidable.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+
+- *What aspects of the design do you expect to clarify further through the RFC review process?*
+- *What aspects of the design do you expect to clarify later during iterative development of this RFC?*
+
+### Drawbacks and Limitations [optional]
+
+* This introduces the ability to permanently destroy data from the DSS. 
+* There is no limit on the amount of data that can be rapidly permanently deleted.
+* Permanently deletion does not occur until after the grace period has elapsed. This could
+  be problematic if files need to be removed sooner.
```
justincc: The wranglers say that the general expectation based on other archiving services is that when it comes to consent issues is to have complete removal in 24-48 hours. They are going to discuss this further at the wrangler F2F on Mon 4th Feb. For redaction not linked to consent it's okay to take weeks.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+   <uuid>.<version>
+   .
+   .
+   .
+
+   confirm (yes or no)?
+
+   ```
+
+3) **Logical Deletion:** The DSS places a **bundle tombstone** in the
```
justincc: Have you considered the case when a caller may require logical deletion but not necessarily redaction? We have identified a few cases [here](https://docs.google.com/document/d/1XeAw05RAJe20Q1XDF7eLWLGF3gCme4-8l-ZfhtDWOLg/edit?ts=5beda8b0#). It might be that in the implementation there is no distinction required - it's not expensive (in fact cheaper in terms of storage costs) for the datastore to eventually entirely remove everything than just the route to retrieve the data? Possibly still worth distinguishing between logical delete and redaction so that just logical delete can be placed on a lower priority entire deletion work queue?

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+Establish a secure and controlled process for deleting data from the Data Store.
+
+
+## Author(s)
+
+*Recommended format for Authors:*
+
+ `[Trent Smith](mailto:tsmith12@ucsc.edu)`
+ 
+ `[Hannes Schmidt](mailto:hannes@ucsc.edu)`
```
mweiden: I'd remove the backticks and just make these links.

rfcs/text/0000-dss-deletion-daemon.md
mweiden: The second clause of a user story should be an action, e.g. "..., I want to remove unconsented data, ..."

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
+### DCP PR:
+
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC number and this pull request in the format:*
```
mweiden: For readability, I'd remove all the instructions from the RFC. You can read them in the RFC template.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+is required. The first section of this specification describes the
+circumstances and mechanisms for logical and physical deletion of files and the
+bundles that contain them.
+
+### User Stories
+
+*Share the [User Stories](https://www.mountaingoatsoftware.com/agile/user-stories) motivating this RFC.*
+As an operator of the DSS, I found unconsented data in the DSS that must be removed, to protect user's privacy.
+
+As a user of the DSS, I would like to know what bundles have been deleted, so I can keep my personal index up-to-date.
```
mweiden: To @justincc's point about FAIR principles, I'd split this story into two:

* As a user of the DSS, I would like notifications of bundle deletion, so that I can keep my personal projections of the data up-to-date.
* As a user of the DSS, I would like the API to return a status indicating bundle deletion, so that I know what happened to a bundle that I used previously.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+*Share the [User Stories](https://www.mountaingoatsoftware.com/agile/user-stories) motivating this RFC.*
+As an operator of the DSS, I found unconsented data in the DSS that must be removed, to protect user's privacy.
+
+As a user of the DSS, I would like to know what bundles have been deleted, so I can keep my personal index up-to-date. 
+
+## Scientific "guardrails" [optional]
+
+*Describe recommended or mandated review from HCA Science governance to ensure that the RFC addresses the needs of the scientific community.*
+
+## Detailed Design
```
mweiden: ?

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+
+3) **Logical Deletion:** The DSS places a **bundle tombstone** in the
+   underlying storage (S3 or GCP bucket). The tombstone's content is as follows:
+   ```JSON
+   { "reason": "<reason>",
+     "details": "<additional info>",
+     "admin_email": "<admin>"
+   }
+   ```
+   The tombstone causes `GET /bundle/{uuid}` to return a 404. If `{version}`
```
mweiden: The description here of how tombstones work and how the API behaves is not quite accurate (see https://github.com/HumanCellAtlas/data-store/blob/a3743a13ed27149d83411ca964e2bea41f3abeec/dss/storage/identifiers.py#L88-L89 and the bundle api). Please check in with me if you have further questions on this.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+     "admin_email": "<admin>"
+   }
+   ```
+   The tombstone causes `GET /bundle/{uuid}` to return a 404. If `{version}` 
+   signifies the latest version of the bundle
+   identified by `{uuid}`, requests for `GET /bundle/{uuid}` without a version
+   also return a 404. The same applies to HEAD requests, both versioned and
+   unversioned. The file versions referenced by the deleted bundle version are
+   not immediately affected by the deletion and remain accessible. The bundle and associated files are added to 
+   a deletion queue for regularly schedule **Physical Deletion**.
```
mweiden: We should discuss physical deletion being continent on specific reasons of a logical ("soft") delete. See https://landing.google.com/sre/sre-book/chapters/data-integrity/#how-google-sre-faces-the-challenges-of-data-integrity-G1sYCe

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+   
+   A notification is sent out to subscribers of datastore deletions when a new bundle tombstone is created, so they can 
+   update their index appropriately.
+
```
mweiden: I would add a bit in here about the `GET` parts of the API returning "deleted" statuses for data that has been deleted.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+1) An administrator identifies the Data Store file versions to be deleted. The
+   administrator then identifies the Data Store bundle versions referencing
+   those file versions.
+
+2) The administrator runs the admin CLI tool to queue the bundle for deletion using the bundles UUID and version.
+   The bundle UUID and version is required for deletion. The admin CLI will return a list of files that will be 
+   deleted and bundles that are effected by this deletion. Bundles may be affected if they share files or are derived 
+   from the deleted bundle.
+   
+   ```bash
```
mweiden: I'd just name the tool and where it is. This CLI output is likely to fall out of date quickly and doesn't add too much signal.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+discretion of the administrator performing the deletion. That discretion is
+guided by a standard operating procedure (SOP).
+
+### Process
+
+Once it is determined that a (meta)data file needs to be physically deleted –
+the details of this determination are the scope of an SOP – the following
+procedure takes place:
+
+1) An administrator identifies the Data Store file versions to be deleted. The
```
mweiden: This process only applies to "admin deletes" from developers and operators of the DSS. If we make data deletable from the API, you'll need to differentiate the cases.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+   A physical delete makes all files associated with a bundle or bundle version inaccessible. This will make HEAD and 
+   GET requests against those file versions return a 404.
+   
+   A log entry is produced when ever the deletion daemon is run. It contains information on what bundles and files
+   have been logically deleted, the reason, the admin who executed the deletion, and info about the deletion markers. 
+   The same information is stored in the **Deletion Table**. The deletion table tracks all bundles and files currently
+   pending permanent deletion. The Deletion Table contains enough information to reverse Physical Deletes before 
+   the data is Permanently deleted.
+
+5) Data that has been physical deleted remain in the bucket for 7 days before being permanently deleted from the bucket.
```
mweiden: 👍

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+* [AWS Object Lifecycle Management](https://docs.aws.amazon.com/AmazonS3/latest/dev/object-lifecycle-mgmt.html),
+* [Amazon DynamoDB TTL](https://aws.amazon.com/about-aws/whats-new/2017/02/amazon-dynamodb-now-supports-automatic-item-expiration-with-time-to-live-ttl/)
+* [GCS Object Versioning](https://cloud.google.com/storage/docs/object-versioning),
+* [GCS Object Lifecycle Management](https://cloud.google.com/storage/docs/lifecycle)
+* [HumanCellAtlas/data-store#1662 Explore Bucket Versioning for Deletion](https://github.com/HumanCellAtlas/data-store/issues/1662)
+* [HumanCellAtlas/data-store#1663 Deletion Daemon Design](https://github.com/HumanCellAtlas/data-store/issues/1663)
+
+### Open Issues
+
+TODO: remove support for `all-versions` tombstone in Data Store
```
mweiden: What's the reasoning here?

rfcs/text/0000-dss-deletion-daemon.md
kislyuk: > Deletion in general should never happen.

I don't think this is a practical rule to follow. There will be normal, non-emergency situations where deletion is wanted.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+
+TODO: implement object deletion and restoration for AWS anf GCP
+
+TODO: implement delete_bundle.py
+
+TODO: implement restore_bundle.py
+
+TODO: enable bucket versioning and lifecycle rules for replicas.
+
+TODO: Remove `Bundle/Delete` API from DSS.
```
mweiden: We should definitely keep deletion functionality in the DSS. As we scale, wranglers will need self-service delete.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+
+### Unresolved Questions
+
+- *What aspects of the design do you expect to clarify further through the RFC review process?*
+- *What aspects of the design do you expect to clarify later during iterative development of this RFC?*
+
+### Drawbacks and Limitations [optional]
+
+* This introduces the ability to permanently destroy data from the DSS. 
+* There is no limit on the amount of data that can be rapidly permanently deleted.
```
mweiden: We should put in a rate limiter.

mweiden (approved):

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+guided by a standard operating procedure (SOP).
+
+### Process
+
+Once it is determined that a (meta)data file needs to be physically deleted –
+the details of this determination are the scope of an SOP – the following
+procedure takes place:
+
+1) An administrator identifies the Data Store file versions to be deleted. The
+   administrator then identifies the Data Store bundle versions referencing
```
kislyuk: Who is an "administrator" per this RFC?

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+   those file versions.
+
+2) The administrator runs the admin CLI tool to queue the bundle for deletion using the bundles UUID and version.
```
kislyuk: What is the "admin CLI tool"? I think the deletion API should be part of the normal DCP CLI, and not any special admin system. Data deletion is an API concept, it should not be done through an admin backdoor.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+   The bundle UUID and version is required for deletion. The admin CLI will return a list of files that will be 
+   deleted and bundles that are effected by this deletion. Bundles may be affected if they share files or are derived 
+   from the deleted bundle.
+   
+   ```bash
+   delete_bundle.py <uuid>.<version> --reason [consent_withdrawn, consent_absent, service_disruption, legal] 
+    --type [logical, physical] --details "<additional info>"
```
kislyuk: Is this meant to be an exhaustive list of possible reasons? I ask because I don't think the DSS should impose any constraints on the reason for deletion. New reasons may arise.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+   
+   A log entry is produced when ever the deletion daemon is run. It contains information on what bundles and files
+   have been logically deleted, the reason, the admin who executed the deletion, and info about the deletion markers. 
+   The same information is stored in the **Deletion Table**. The deletion table tracks all bundles and files currently
+   pending permanent deletion. The Deletion Table contains enough information to reverse Physical Deletes before 
+   the data is Permanently deleted.
+
+5) Data that has been physical deleted remain in the bucket for 7 days before being permanently deleted from the bucket.
+   This grace period allows the deletion to be reversed within that time period. After the grace period the data will 
+   be permanently deleted.
```
kislyuk: I don't believe a deletion bucket is necessary. Instead, data that is scheduled for physical deletion should be made inaccessible (logical deletion) and remain in the main bucket for a grace period.

Introducing a deletion bucket forces us to unarchive and copy data in order to delete it. That process itself can be long and cumbersome, and I don't believe the benefits of having a deletion bucket outweigh the problems with it.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+It is the administrator's responsibility to enumerate all bundles referring to
+a file that is to be deleted and issue a delete requests for each of
+them. If the administrator misses such a bundle (a *dangling bundle*), users
+requesting `GET /bundle` for a dangling bundle will be able to retrieve its
+manifest. Likewise, `POST /search` requests will return the metadata for
+dangling bundles. Secondary indexes like the Data Browser (Orange Box) will
+continue to hand out (meta)data for dangling bundles. If users request `GET
+/files` for a file contained in a dangling bundle and the requested file was
+referenced by another already logically (physically) deleted bundle, the Data
+Store will respond with a 404 (500) status code.
```
kislyuk: Is there currently a way to list all bundles that a file is part of in the API? If not, we should consider adding one.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+The metadata in bundles containing derived data must reference the input bundle
+versions. Administrators are responsible for tracking down data bundles with
+data derived from data in deleted bundles and to issue deletion requests for
+them.
+
+### Tiered storage
+
+Physical deletion of blobs is effective across all tiered storage classes in
+the underlying blob store. Additional cost may be incurred for deleting data recently moved to a
+slower access storage tier. Deletion occur There is no additional delays cause when deleting from a
```
kislyuk: Editing typo here (" Deletion occur ")

mweiden (changes requested):
Meant for my last review to be a "Request changes"

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+### Tiered storage
+
+Physical deletion of blobs is effective across all tiered storage classes in
+the underlying blob store. Additional cost may be incurred for deleting data recently moved to a
+slower access storage tier. Deletion occur There is no additional delays cause when deleting from a 
+tiered storage system such as AWS glacier.
+
+### Undoing logical deletions
+
+A bundle can be restored using `restore_bundle.py <uuid.<version>` if has not been physically deleted. This process
```
kislyuk: This is a DCP service RFC, and as such, I think it should specify the semantics of HTTP REST/RPC APIs, not details of Python scripts. The deletions semantics described in this RFC should be available in an API, not through a DSS service admin "backdoor" defined by a Python script.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,268 @@
...
+uses the info from the deletion table to removes tombstones, and prevent permanent deletion. Subscribers of new 
+bundles will receive a notification that a new bundle has arrived.
+
+Deletion Table Entry Example:
+
+|Bundle.Version|admin|reason|AWS Deletion Markers (key,ID)|GCP Previous Generations (key, previous generation)|
+|--------------|-----|------|--------------------|------------------------|
+|1234-2134-3145|admin@email.com|consent|(file.obj, 1234), ...| (file.obj, 4321), ...|
```
kislyuk: Do I understand correctly that the deletion table is an *implementation detail* of the DSS, internal to DSS and not exposed in the API? I think it's OK and appropriate to list the details of it here, as long as that fact is called out. More generally, throughout this RFC, more clarity is necessary to separate the background information, the API for the deletion actions, and the implementation details of the deletion service infrastructure.

rfcs/text/0000-dss-deletion-daemon.md
kislyuk: To add to @mweiden's comment: As stated in my other comments here, I think the deletion system should have a well-defined API, and that API should be served over HTTP and available through the same endpoint as the rest of DSS. It should not be a maintenance script requiring infrastructure access. It should be secured using the shared auth infrastructure that we developed partly for this purpose.

rfcs/text/0000-dss-deletion-daemon.md
kislyuk: This can be addressed by a suite of safety checks that could throttle deletions in the deletion daemon based on size delta, object count delta, etc. and send alerts to the DCP operators.

rfcs/text/0000-dss-deletion-daemon.md
kislyuk: @xbrianh I think the RFC should discuss the policies and the API (and the API in this case is a requirement, since there is an action being performed by a service on behalf of a user who is not the service operator). I think the RFC *can* go into detail about the implementation, as long as such discussion is clearly marked as an informational implementation detail, not a commitment to API semantics or policy.

kislyuk (changes requested):
I don't think we should strive to build a system that relies on data wranglers passing ad hoc information to service operators, and service operators running maintenance scripts, for an operation as core to the data lifecycle as deletion. Such a system is error-prone, non-scalable, and stifles data operations by keeping data management tools out of the hands of data operators.

More generally, I think we should re-examine our architectural assumptions and principles here, and consider how this RFC corresponds to the core DCP principle of federated services with boundaries defined by public, documented HTTP APIs. This principle is key to our agility.

rfcs/text/0000-dss-deletion-daemon.md
Bento007: @kislyuk but you agree with the presented reason for deleting data?

rfcs/text/0000-dss-deletion-daemon.md
Bento007: not exhaustive and I image will expand in the future. I believe by forcing the admin to consider the specific reason for deletion a powerful tool in prevent excessive deletions. We could simply use `--details` but the free formed nature of that field makes it easier to abuse. Not that any of use would do that. Obviously this is not set in stone and can be changed.

rfcs/text/0000-dss-deletion-daemon.md
Bento007: this would be an extension to the existing [admin-cli.py](https://github.com/HumanCellAtlas/data-store/blob/master/scripts/admin-cli.py)

rfcs/text/0000-dss-deletion-daemon.md
Bento007: this RFC is for admin deletes. Allowing users to delete bundles can come later.

Bento007: I think data wrangler as most should be able to perform logical delete and not physical deletes.

rfcs/text/0000-dss-deletion-daemon.md
Bento007: a DCP admin who has access to the admin-cli. They would have access to the AWS and GCP infrastructure.

rfcs/text/0000-dss-deletion-daemon.md
Bento007: this is an obfuscation of the underlying bucket lifecycle policies that will be used to clean up files that have been deleted in a version bucket.

rfcs/text/0000-dss-deletion-daemon.md
Bento007: We can accomplish this with an ES query of the manifest.

rfcs/text/0000-dss-deletion-daemon.md
Bento007: Rather than a rate limiter there will be a mandatory grace period before permanent deletion and notifications sent to admins to inform them about pending deletions. The admin will then have a window of time to reverse any or all deletions.

rfcs/text/0000-dss-deletion-daemon.md
Bento007: Are you referring to the outdate swagger.yml which does not specify a 404 error code? Please elaborate for me.

mweiden: If I recall correctly, this is how `DELETE /v1/bundles` interacts with the other parts of the API

* If you do specify a version on `DELETE /v1/bundles`, only that specific bundle version is tombstoned.
* If you do not specify a version when you `DELETE /v1/bundles` it creates a tombstone without a version. This results in all bundles with that UUID being tombstoned. I believe this applies to future versions of the bundle as well.

rfcs/text/0000-dss-deletion-daemon.md
mweiden: probably the best combo:

1. Grace period
1. Metrics for deletion rate + an alert on high deletion rates

rfcs/text/0000-dss-deletion-daemon.md
Bento007: Fixed

rfcs/text/0000-dss-deletion-daemon.md
Bento007: Added APIs documentation

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,620 @@
...
+
+This is a new addition to the DSS API.
+
+Only **physical deletes** exist for the file deletion API. The file deletion API is a two part process. The first 
+request will return the bundles that will be **logically deleted**
+as a side effect and a confirmation code. The second request must include the confirmation code in order to begin the
+deletion process. The request follows this swagger document:
+
+```yaml
+      /file/{uuid}:
```
mweiden: I understand why you would want to be precise on the specification of the API, though I'm not sure the complete swagger yaml is required to do this... There's a lot of weedy detail here.

Perhaps the query path, short list of parameters, and a simple table of responses would get the point across.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,620 @@
...
+   bundles mentioned in the request response. The tombstone's content is
+   as follows:
+   ```JSON
+   { "reason": "<reason>",
+     "details": "<additional info>",
+     "email": "<requester's email>"
+   }
+   ```
+   
+   A versioned bundle tombstone will cause `GET /bundle/{uuid}` to return a 409 for that bundle version. If the tombstoned
```
mweiden: `HTTP 409` is significant of a "conflict." Why not a `410`?

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,620 @@
...
+   being permanently deleted.
+
+6) Data that has been physical deleted remain in the bucket for 7 days before being permanently deleted from the bucket.
+   This grace period allows the deletion to be reversed within that time period. After the grace period the data will 
+   be permanently deleted.
+ 
+### Tombstones
+
+Tombstones are markers places in the DSS to indicate previously existing data has been removed. Tombstones exist for
+files and bundles stored in the DSS. Tombstones come in two varieties, versioned and unversioned tombstones. A
```
mweiden: `Tombstones exist for files and bundles stored in the DSS.` - This seems to contradict the earlier point about there only being physical deletion for files.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,620 @@
...
+ 
+### Tombstones
+
+Tombstones are markers places in the DSS to indicate previously existing data has been removed. Tombstones exist for
+files and bundles stored in the DSS. Tombstones come in two varieties, versioned and unversioned tombstones. A 
+version tombstone marks the deletion of a specific version of a file or bundle. This prevents that same version of data 
+from being reuploaded or retrieved. An unversioned tombstone does not specify a version and prevent a file or bundle 
+using the same UUID from being reuploaded or retrieved. Whether a tombstone is version or unversioned is determined when
+during a deletion request. If the version parameter is present in the request, then a versioned tombstone is created. If
+no version is specified then an unversion tombstone is created. A versioned tombstone is created for all bundles and
```
mweiden: This is a little misleading. If no version is specified, actually an _unversioned_ tombstone is created that is applied to all the bundles with that UUID.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,620 @@
...
+   the data is Permanently deleted.
+   
+   A daily digest of bundles and files that are scheduled for deletion are sent to DCP administrator 24 hours prior to the data
+   being permanently deleted.
+
+6) Data that has been physical deleted remain in the bucket for 7 days before being permanently deleted from the bucket.
+   This grace period allows the deletion to be reversed within that time period. After the grace period the data will 
+   be permanently deleted.
+ 
+### Tombstones
```
mweiden: Tombstones are really an implementation detail of logical deletion within DSS. I'm not sure that external users really need to know about them.

rfcs/text/0000-dss-deletion-daemon.md
Bento007: your right this should be 410. Will fix

rfcs/text/0000-dss-deletion-daemon.md
Bento007: Added section on monitoring

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,631 @@
...
+   A physical delete makes all files associated with a bundle or bundle version inaccessible. This will cause direct
+   urls to a deleted file to fail.
+   
+   A log entry is produced when ever the deletion daemon is run. It contains information on what bundles and files
+   have been logically deleted, the reason, the administrator who executed the deletion, and info about the deletion markers. 
+   The same information is stored in the **Deletion Table**. The deletion table tracks all bundles and files currently
+   pending permanent deletion. The Deletion Table contains enough information to reverse Physical Deletes before 
+   the data is Permanently deleted.
+   
+   A daily digest of bundles and files that are scheduled for deletion are sent to DCP administrator 24 hours prior to the data
```
DailyDreaming: Would it be more useful to just send a digest when things are added rather than on a 24 hour timer?

Things are always going to be added to the queue less than 24 hours prior to deletion if the deletion occurs every 24 hours.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,631 @@
...
+    
+A user must have explicit permission to perform a **logical deletion** of a bundle. For a **physical deletion** of a
+bundle, the user must have explicit permission to **physically delete** files and bundles.
+
+(!) The deletion of a bundle does not handle the deletion of secondary analysis bundles referencing this bundle via the 
+links metadata field.
+
+### Restore File API
+
+This is a new endpoint `PUT /restore/file/{uuid}` added to the DSS API which follows this swagger:
```
DailyDreaming: This endpoint seems to have 2 clear use cases from `### Undoing logical deletions`:

1. Restoring a file that is still within the 7 day grace period.
2. Restoring a file that was permanently deleted.

I'm not sure this swagger seems to have params for each of these cases.  For `2.` you would need to specify where it's getting the file from.

Bento007: The restore file only handle case 1. Case 2 would require re ingesting the file with a new file version or file uuid.version

Bento007: Does this address your comment https://github.com/HumanCellAtlas/dcp-community/pull/57/commits/c20a038b70d012746892f5334bf3cae2acee81f0

rfcs/text/0000-dss-deletion-daemon.md
Bento007: There are 3 stage of deletion in total. Only two stages of deletion that Users care about. Logical deletion basically the creation of a tombstone and making it invisible to API. Physical deletion is using the bucket versioning to delete the item from the bucket without completely removing it. Physical and Logical are the only ones uses should worry about. Permanent deletion uses the bucket life cycle policies to clean up deleted files once they are so old. The time between physical deletion and permanent deletion is the grace period. We know when the physical deletion occurred and how long the grace period is we can determine how long before the physically deleted data. When <24 hours remains before the grace period is up we'd like to send a notification to warn about pending permanent deletions.

<36 hours might actually be a better time window for notifications.

Bento007: I think it would get spammy if we send a digest every time a deletion occurred when it should be impossible to permanently delete something earlier then the grace period.

DailyDreaming: Ah, that clears it up, it's just confusion since the term deletion is being used interchangeably for all 3 steps at times.  Yeah, that seems to address it.

rfcs/text/0000-dss-deletion-daemon.md
DailyDreaming: Yes, that certainly takes care of it.  Though for case 2, would that just be the normal process of adding a new file or something special?

Bento007: It would be a normal re ingest of the data.

rfcs/text/0000-dss-deletion-daemon.md
kislyuk: Yes.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,631 @@
...
+   the deletion queue for regularly schedule **Physical Deletion**. A **physical deletion** request of a file result in 
+   that file being added to the to deletion queue
+
+5) On a regular schedule the deletion queue is processed by code running within a secure boundary. An administrator 
+   may manually invoke the deletion process or wait for it run at it's scheduled time once daily.
+   The deletion daemon is responsible for performing **physical deletes** of data from the deletion queue. 
+   A physical delete makes all files associated with a bundle or bundle version inaccessible. This will cause direct
+   urls to a deleted file to fail.
+   
+   A log entry is produced when ever the deletion daemon is run. It contains information on what bundles and files
```
amarjandu: Would information regarding the user that created the delete request show within the `Get Deleted File API` line 433?

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,631 @@
...
+                  properties:
+                    confirmation:
+                      type: string
+                      description: a key used to confirm the deletion operation
+            201:
+              description: the deleted bundle and associated file restoration process has been confirmed.
+            403:
+              description: Unauthorized user it attempting this action.
+            404:
+              description: the bundle cannot be restored because it was never deleted, or was logically deleted. This can
```
amarjandu: should this read, "or was physically deleted", referencing section starting on line 566

Bento007: If it were physically deleted then restoration would not return an error.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+|200| Deletion pending. A confirmation code is returned to confirm the deletion in the second request. A list of affect files and bundles is also returned |
+|201| Deletion Confirmed with confirmation code. |
+|403| Unauthorized user is attempting this action.|
+|404| The bundle has already been deleted.|
+|409| The confirmation code used is invalid. Either the confirmation code was incorrect, or an effected bundle has changed state since the original request.|
+    
+A user must have explicit permission to perform a **logical deletion** of a bundle. For a **physical deletion** of a
+bundle, the user must have explicit permission to **physically delete** files and bundles.
+
+(!) The deletion of a bundle does not handle the deletion of secondary analysis bundles referencing this bundle via the
```
justincc: @parthshahva you are considering this for redaction design?

rfcs/text/0000-dss-deletion-daemon.md
lauraclarke: Based on the presentations from Dave Bernick from a FISMA compliance perspective speed of deletion doesn't need to be so quick. That being said from a reputational perspective (and considering any physical deletion request will need to navigate a certain amount of bureaucracy before the technical request is made) the final step taking 24 to 48hrs seems like a good aim

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+Under certain circumstances however, the **physical deletion** of a file version
+is required. The first section of this specification describes the
+circumstances and mechanisms for logical and physical deletion of files and the
+bundles containing them. The second section details the changes to the API required to manage the deletion process.
+The final section details the process of deleting data from the DSS, and design details.
+
+### User Stories
+
+* As a user of the DSS, I would like to know what bundles have been deleted, so I can keep my personal index up-to-date. 
+* As a data wrangler of the DSS, I need to remove data from the DSS, because there was no consent from the donor to share the data.
```
lauraclarke: From a strategic viewpoint, this user story should be written from the perspective of the user informing us of the consent issues. The Wrangler is almost always only an messenger.

I would suggest

As a data contributor, I need to remove data from the DSS as I have realised the consent it was collected under does not allow it to be hosted by the DCP.

mweiden (approved):

mweiden:
@Bento007 the issues I saw have been resolved, so I'm approving. The others have raised some good points, so I'd continue to work through those.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+
+Under certain circumstances however, the **physical deletion** of a file version
+is required. The first section of this specification describes the
+circumstances and mechanisms for logical and physical deletion of files and the
+bundles containing them. The second section details the changes to the API required to manage the deletion process.
+The final section details the process of deleting data from the DSS, and design details.
+
+### User Stories
+
+* As a secondary index of the DSS, I would like to know what bundles have been deleted, so I can keep my personal index up-to-date.
```
kislyuk: "As an operator of a secondary index" perhaps?

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+  
+* storing or processing the (meta)data causes a serious disruption of service
+  within the DCP
+
+* the (meta)data was submitted to the DCP by a malicious party in violation of
+  the DCP's data sharing agreement. This includes (meta)data that violates the
+  laws or regulations in a jurisdiction applicable to the DCP or the (meta)data
+  it stores
+
+This is not an exhaustive list, and more reason maybe discovered later.
```
kislyuk: "more reasons may be"

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+|201| Deletion Confirmed with confirmation code. |
+|403| Unauthorized user it attempting this action.|
+|404| The files has already been deleted.|
+|409| The confimation code used is invalid. Either the confirmation code was incorrect, or an effected bundle has changed state since the original request.|
+
+A user must have explicit permission to perform the file deletion operation.
+
+### Bundle Deletion API
+
+The bundle deletion API is a two part process. The first request will return the bundle and files that will be affect
```
kislyuk: "will be affected"

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+   ```JSON
+   { "reason": "<reason>",
+     "details": "<additional info>",
+     "email": "<requester's email>"
+   }
+   ```
+   
+   A versioned bundle tombstone will cause `GET /bundle/{uuid}` to return a 410 for that bundle version. If the tombstoned
+   bundle is the latest version then a request for `GET /bundle/{uuid}` without a version will
+   also returns a 410. If bundle tombstone does not specify a bundle version then all request for `GET /bundle/{uuid}` will
```
kislyuk: s/returns/return/
s/request/requests/

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+   bundles mentioned in the request response. The tombstone's content is
+   as follows:
+   ```JSON
+   { "reason": "<reason>",
+     "details": "<additional info>",
+     "email": "<requester's email>"
+   }
+   ```
+   
+   A versioned bundle tombstone will cause `GET /bundle/{uuid}` to return a 410 for that bundle version. If the tombstoned
```
kislyuk: Here and elsewhere, formally it should be "return a 410 status code"

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+   bundle is the latest version then a request for `GET /bundle/{uuid}` without a version will
+   also returns a 410. If bundle tombstone does not specify a bundle version then all request for `GET /bundle/{uuid}` will
+   return 410 even if version is specified in the request. The same applies to `HEAD /bundle/{uuid}` requests, both versioned and
+   unversioned. The same applied for `GET /files/{uuid}` and `HEAD /files/{uuid}`. Bundles that have an
```
kislyuk: applies

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+   }
+   ```
+   
+   A versioned bundle tombstone will cause `GET /bundle/{uuid}` to return a 410 for that bundle version. If the tombstoned
+   bundle is the latest version then a request for `GET /bundle/{uuid}` without a version will
+   also returns a 410. If bundle tombstone does not specify a bundle version then all request for `GET /bundle/{uuid}` will
+   return 410 even if version is specified in the request. The same applies to `HEAD /bundle/{uuid}` requests, both versioned and
+   unversioned. The same applied for `GET /files/{uuid}` and `HEAD /files/{uuid}`. Bundles that have an 
+   associated tombstone are inaccessible through the API and are considered **logically deleted**. Files with an 
+   associated tombstone or not logically delete. User with a direct URL to the file can still access the data.
```
kislyuk: "or not logically delete" -> ??? Not sure exactly but that sentence does not parse.

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+   
+   A notification is sent out to subscribers of datastore deletions when a new bundle tombstone is created, so they can 
+   update their index appropriately. If the the bundle was logically deleted and later physically deleted a second
+   identical notification will be sent to subscribers. The subscribers must be able to handle these notifications 
+   idempotently.
+   
+   A **physical deletion** request of a bundles results in the bundle associated files being added to
```
kislyuk: s/bundles/bundle

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+   associated tombstone are inaccessible through the API and are considered **logically deleted**. Files with an 
+   associated tombstone or not logically delete. User with a direct URL to the file can still access the data.
+   
+   A notification is sent out to subscribers of datastore deletions when a new bundle tombstone is created, so they can 
+   update their index appropriately. If the the bundle was logically deleted and later physically deleted a second
+   identical notification will be sent to subscribers. The subscribers must be able to handle these notifications 
+   idempotently.
+   
+   A **physical deletion** request of a bundles results in the bundle associated files being added to 
+   the deletion queue for regularly schedule **Physical Deletion**. A **physical deletion** request of a file result in
```
kislyuk: s/schedule/scheduled

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+   update their index appropriately. If the the bundle was logically deleted and later physically deleted a second
+   identical notification will be sent to subscribers. The subscribers must be able to handle these notifications 
+   idempotently.
+   
+   A **physical deletion** request of a bundles results in the bundle associated files being added to 
+   the deletion queue for regularly schedule **Physical Deletion**. A **physical deletion** request of a file result in 
+   that file being added to the to deletion queue
+
+5) On a regular schedule the deletion queue is processed by code running within a secure boundary. An administrator 
+   may manually invoke the deletion process or wait for it run at it's scheduled time once daily.
```
kislyuk: "wait for it to run"
"it's"->"its"

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+Elasticsearch's dynamic mapping is enabled.
+
+### Replication
+
+When a bundle or file is deleted it is removed from all replicas.
+
+### Consistency
+
+Race conditions with Elasticsearch can arise when deleting files or bundles. The confirmation code is used to ensure
+that the state of the index has not changes between deletion requests. If the state has changed another request to delete
```
kislyuk: changed

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+The metadata in bundles containing derived data must reference the input bundle
+versions. Administrators are responsible for tracking down data bundles with
+data derived from data in deleted bundles and to issue deletion requests for
+them.
+
+### Tiered storage
+
+Physical deletion of blobs is effective across all tiered storage classes in
+the underlying blob store. Additional cost may be incurred for deleting data recently moved to a
+slower access storage tier. There is no additional delays cause when deleting from a
```
kislyuk: There is no additional delays cause -> ?

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+### Tiered storage
+
+Physical deletion of blobs is effective across all tiered storage classes in
+the underlying blob store. Additional cost may be incurred for deleting data recently moved to a
+slower access storage tier. There is no additional delays cause when deleting from a 
+tiered storage system such as AWS glacier.
+
+### Undoing Deletions
+
+A bundles can be restored using `PUT restore/bundle/{uuid}` if bundle has been logically
```
kislyuk: "A bundle"

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+tiered storage system such as AWS glacier.
+
+### Undoing Deletions
+
+A bundles can be restored using `PUT restore/bundle/{uuid}` if bundle has been logically 
+deleted. Physically deleted bundles and files can be restored using the restore API if the grace period for the data 
+has not elapsed. Subscribers of bundles will receive a notification that a new bundle has arrived when a bundle is restored.
+
+Within the DSS the process for restoring data uses the info from the deletion table to removes tombstones, and prevent 
+permanent deletion.cDeletion Table Entry Example:
```
kislyuk: Typo here

rfcs/text/0000-dss-deletion-daemon.md
```diff
@@ -0,0 +1,416 @@
...
+### Open Issues
+
+* TODO: implement object deletion and restoration for AWS and GCP
+* TODO: implement tombstones for files
+* TODO: enable bucket versioning and lifecycle rules for replicas.
+
+### Acceptance Criteria [optional]
+
+* When a bundle is tombstoned it is made immediately unavailable to users using search.
+* Maintainers of external indices shall removed deleted data from its index when a bundle tombstone notification is received.
```
kislyuk: "shall remove"

kislyuk (approved):
LGTM overall; my major concerns have been addressed. Great job @Bento007.

I suggest proofreading the doc.

rfcs/text/0000-dss-deletion-daemon.md
Bento007: will fix

rfcs/text/0000-dss-deletion-daemon.md
Bento007: will fix

rfcs/text/0000-dss-deletion-daemon.md
Bento007: will fix

rfcs/text/0000-dss-deletion-daemon.md
Bento007: will fix

rfcs/text/0000-dss-deletion-daemon.md
Bento007: will fix

rfcs/text/0000-dss-deletion-daemon.md
Bento007: will fix

rfcs/text/0000-dss-deletion-daemon.md
Bento007: Fixed: changed to "Files with an associated tombstone are not considered logically deleted, because user with a direct URL to the file can still access the data."

rfcs/text/0000-dss-deletion-daemon.md
Bento007: Fixed

rfcs/text/0000-dss-deletion-daemon.md
Bento007: Fixed

rfcs/text/0000-dss-deletion-daemon.md
Bento007: fixed

rfcs/text/0000-dss-deletion-daemon.md
Bento007: changed to "There is no additional time delay caused when deleting from a tiered storage system such as AWS glacier."

rfcs/text/0000-dss-deletion-daemon.md
Bento007: fixed

rfcs/text/0000-dss-deletion-daemon.md
Bento007: fixed

rfcs/text/0000-dss-deletion-daemon.md
Bento007: Fixed

rfcs/text/0000-dss-deletion-process.md
```diff
@@ -0,0 +1,416 @@
...
+|--------------|------------|
+|410| The file has been deleted. The reason and details for the deletion are returned. |
+
+**Path**: `GET /file/{uuid}`
+
+|Response Code| Description|
+|--------------|------------|
+|410| The file has been deleted. The reason and details for the deletion are returned. |
+
+**Path**: `PUT /bundles/{uuid}`
```
samanehsan: Did you mean `PUT /files/{uuid}` here?

rfcs/text/0000-dss-deletion-process.md
```diff
@@ -0,0 +1,416 @@
...
+ [Hannes Schmidt](mailto:hannes@ucsc.edu)
+
+## Shepherd
+ [Rhian Anthony](mailto:ranthony@broadinstitute.org)
+
+## Motivation
+
+One of the DCP's primary objectives is to ensure the durability of the
+(meta)data it contains and to maintain a complete record of the changes made to
+it. The DCP achieves this by never overwriting a (meta)data file but instead
```
sampierson: AFAIK this is not a primary objective of the DCP - please provide a reference to this requirement if you have one.  I don't think people care that we record the 5 broken versions of a bundle that have incorrect metadata, so much as having a persistent way to refer to and store the correct version(s) of bundles.

I think it _may_ be a design goal of the DSS, but is also possibly just a simplifying assumption for the DSS implementation.

rfcs/text/0000-dss-deletion-process.md
Bento007: Yes, it should be file. Will fix

rfcs/text/0000-dss-deletion-process.md
Bento007: Do you have a recommendation or would it be better to remove all together?

kislyuk: @sampierson while not a "primary objective", I think it's transitively a stated shared goal of the DCP because the DSS owns the responsibility and process around storing and providing access to data (https://github.com/HumanCellAtlas/dcp-community/blob/master/charters/DataStore/charter.md). In order to fulfill this responsibility, in particular in order to provide both data revision ("AUDR") capabilities and stable (immutable) release URI capabilities, the DSS implements an immutable versioned storage model.

PR #58, opened 2019-01-24
Add Integration testing for metadata schema changes RFC

mweiden:
~Last call for community review: 2019-02-07~
Last call for oversight review: 2019-03-01

High level summary of review discussions:

@simonjupp @tburdett @daniwelter suggested that strategies should be considered for selecting a sample set of data so that they exercise new schema changes in the integration test. @mweiden added text that there should be a selection strategy for bundle sampling.

@simonjupp @malloryfreeberg @tburdett raised concerns about coupling of components in integrations test. Specifically failing downstream paths shouldn’t prevent upstream paths from passing. @mweiden addressed this concern by proposing that DSS subscription API should enable clients to opt into receiving events with unsupported schema versions. 

@tburdett made a number of suggestions clarifying the use of language that were addressed.


rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,78 @@
...
+ 
+ * "schema integration test system" - an automated CI/CD pipeline used to check that downstream systems can handle bundles with a new schema
+ * "new schema label" - anything that indicates that the schema has changes that have not yet been tested against downstream systems: potentially a JSON field, git tag, git branch, or comparison to tested version list
+
+
+Design
+
+1. Bundles with new metadata schema changes are uploaded to the production environment with a new schema label
+1. Observing the new schema label, the DSS stores the new bundles and makes them available with the `GET /v1/bundles` endpoint with the bundle `uuid` or `(uuid, version)`, but does not index nor does it release subscription notifications for them.
+1. Via a schema integration test system, a sample of the uploaded bundles are copied from the production DSS to the integration DSS testing environment.
```
simonjupp: It's possible a random sample might not exercise any of the new schema changes

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,78 @@
...
+* Bundles only have to be uploaded once
+* Bundles can be uploaded directly to the production environment
+* Tests exactly what bundles downstream systems will see; does not require complicated fake metadata generation code in tests
+* Data in storage is immediately available after upload via the `GET /v1/bundles` endpoint
+* Downstream systems are protected from new schema changes until it can be confirmed that they are able to handle the changes
+* Does not require replaying subscription notifications
+
+### Cons:
+
+* Does not test new bundles against the ingestion pipeline, this is assumed to "just work"
```
simonjupp: what does this mean? ingest doesn't ingest bundles.

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,78 @@
...
+1. Via a schema integration test system, a sample of the uploaded bundles are copied from the production DSS to the integration DSS testing environment.
+1. The integration DSS deployment stores the bundles, indexes their metadata, and releases bundle events to downstream systems in the integration environment with subscriptions.
+1. The schema integration test checks the results and passes if systems downstream of the integration DSS have correctly processed the new bundles; it fails otherwise.*
+1. If the schema integration test passes, the new bundles in the production DSS can be released downstream by issuing a new bundle version of each without the new schema label. This process could be automated.
+1. If the schema test does not pass, the data is not released and development teams are notified to resolve the issue and rerun the test.
+
+![Integration test design](../images/0000-integration-test.png)
+
+*Integration test data flow: test path is shown in red, production release in blue, integration tooling in dashed lines; edges are labeled with corresponding step numbers in the design above*
```
simonjupp: If the analysis passes but the browser fails, will the data be released in production for analysis pipelines?

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Cons:
+
+* Does not test new bundles against the ingestion pipeline, this is assumed to "just work"
+* Does not cover third party systems, which may be consuming from the DSS
+* New bundles are not available in the data browser until tests for the entire system have passed
+* There is no hard guarantee that downstream production systems will be able to handle new bundles: there could be critical updates in the integration environment which have not yet reached production
+* In the case that the integration test does not pass, developers and wranglers will have to exercise care in propagating fixes to production before releasing the new bundles downstream of the production DSS
+
+## Discussion:
```
simonjupp: Overall I think checking a sample of metadata is not sufficient. Downstream consumers can check their assumptions against the schema itself

rfcs/text/0000-metadata-integration-test.md
mweiden: Good point. What strategies for selecting the sample would you suggest?

The nice thing about the approach is that it tests the actual data that must pass through the system. The problems raised would be exactly those we would need to solve in order to deliver the final data product to our users.

rfcs/text/0000-metadata-integration-test.md
mweiden: What we mean here is that the first pass of the data through ingest and upload (step 1) is not part of the formal integration test, but rather a live run of the production system.

Suggested rewording?

rfcs/text/0000-metadata-integration-test.md
mweiden: In this design, no. The rationale:

1. Once the data is released and events are sent out, it is tricky to track and replay those that have fallen out of the retry window (though this window is long). I don't think we want to get into the weeds of tracking which clients have missed which events and replaying them indefinitely. 
1. There are some systems, e.g. Lira+Pipelines, that might accept an event successfully, but fail further downstream in an internal component. In this case, DSS cannot know that it needs to replay the event to them. Unless there is a local replay system or a developer intervenes, this data would not propagate. We should avoid this failure case.

rfcs/text/0000-metadata-integration-test.md
mweiden: Glad you brought this point up. I agree and think validating the client's "contract" on the schema is a subsequent logical step after we have integration testing.

In the future, it would be great to provide users with something like a schema contract library to do this (see [slides](https://docs.google.com/presentation/d/1Y28vT6PbPDC6ublsZIIMNTvianCxVLlt4M5sypwX2wE/edit#slide=id.g4339acaf8a_0_66) and [sample implementation](https://github.com/HumanCellAtlas/metadata-migration#-contracts)).

Importantly, we need proper integration testing too to understand which clients will be broken by a given schema change before we notify data consumers of data published with that schema and to check that our stated dependencies ("contract") on the schema are correct.

rfcs/text/0000-metadata-integration-test.md
simonjupp: I don't think there's a simple strategy for working out what you need to test without just testing all the data. You could select a sample of data for every unique graph shape in the experiment (we're working on tooling to identify different shapes). You might also want to select samples that use specific fields that we know have been either added or changed in the schema, but this would require some level of schema introspection.

rfcs/text/0000-metadata-integration-test.md
simonjupp: How about 

> * Does not test the effect of schema changes on the data contribution pipeline, it is assumed a sibling integration test will exist for contribution of data and metadata as opposed to consumption

rfcs/text/0000-metadata-integration-test.md
simonjupp: I think we need to avoid coupling the consumption by pipelines and consumption by browser.  There will be cases where new types of pipelines will need to run that require newer versions of a particular schema that the browser will not be able to handle. 

If we stick with this design, we should make it clear in the cons that analysed data can't be released to downstream consumers until it passes the browser tests.

rfcs/text/0000-metadata-integration-test.md
simonjupp: +1 to all this

simonjupp (changes requested):
Just changing the review status to request changes while my comments are being worked through

rfcs/text/0000-metadata-integration-test.md
mweiden: Sounds like a great opportunity to apply the tooling you're building!

I Like these ideas. I've put them in as a tentative selection strategy for bundle sampling.

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,80 @@
...
+* [Mark Diekhans](mailto:markd@ucsc.edu)
+* [Hannes Schmidt](mailto:hannes@ucsc.edu)
+
+## Shepherd
+
+[Norman Morrison](mailto:norman@ebi.ac.uk)
+
+## Motivation
+
+Currently breakages to DCP systems caused by metadata schema changes are detected at runtime upon data (bundle) upload. This forces data wranglers, DCP operators, and data consumers to react to unpredictable runtime failures rather than detecting problems before the code is merged and deployed.
```
tburdett: This language in this paragraph is somewhat unclear: bundles are uploaded to the DSS but not the DCP (they are generated by ingest from _submissions_); runtime failures aren't exactly unpredictable (they *could* be anticipated by each component upon release of a new schema, although admittedly the code to do this is somewhat complex); what does "the code" refer to; and why do data wranglers and data consumers need to react (wranglers have successfully handed off new data, consumers will get an error and therefore just fail to incorporate the newest data)?

I'd like to suggest something more like this:

```suggestion
Currently, DCP systems are tightly coupled to specific metadata schema versions and experimental graph shapes, and are therefore susceptible to breakages upon ingestion of previously unseen data. Such breakages are most likely to occur upon the release of a new major version of a metadata schema, although submissions making use of new experimental graph shapes may also trigger a breakage without any schema changes. This forces DCP operators to react to runtime failures caused by attempting to analyse or display this data, and forces data consumers to require frequent updates to code with dependencies on the metadata schema.
```

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,80 @@
...
+
+In this design we will use the following terms:
+ 
+ * "schema integration test system" - an automated CI/CD pipeline used to check that downstream systems can handle bundles with a new schema
+ * "new schema label" - anything that indicates that the schema has changes that have not yet been tested against downstream systems: potentially a JSON field, git tag, git branch, or comparison to tested version list
+
+
+Design
+
+1. Bundles with new metadata schema changes are uploaded to the production environment with a new schema label
```
tburdett: In general I find the use of "bundles" throughout this document quite awkward; it presents a very DSS-centric view of a DCP-wide issue. Can we instead refer to "ingestion of data" here? If you want to keep the bundle focus, "...are pushed to the production Data Store..." (as opposed to "environment") would be fine.

My preferred usage:

- Data files are _uploaded_ to the Upload Service
- Submissions are _ingested_ to the Ingestion Service
- Bundles are _pushed_ to (or data is _stored_ in) the Data Store

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,80 @@
...
+1. Make updates to software and schema smaller and more frequent
+1. Make failures due to metadata schema changes more readily observable
+1. Increase availability of DCP systems downstream of DSS
+
+## Detailed Design
+
+In this design we will use the following terms:
+ 
+ * "schema integration test system" - an automated CI/CD pipeline used to check that downstream systems can handle bundles with a new schema
+ * "new schema label" - anything that indicates that the schema has changes that have not yet been tested against downstream systems: potentially a JSON field, git tag, git branch, or comparison to tested version list
```
tburdett: I think I'm struggling to see what this would look like (other than schema version numbers). How would one know whether a particular submission generated bundles that needed a new schema label? Does this go as far as some sort of hash of all the schema types/versions/fields used in a given set? If so, as @simonjupp mentioned in an earlier comment, this would be extremely likely to be unique for every single submission

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,80 @@
...
+
+![Integration test design](../images/0000-integration-test.png)
+
+*Integration test data flow: test path is shown in red, production release in blue, integration tooling in dashed lines; edges are labeled with corresponding step numbers in the design above*
+
+\* To estimate the impact of schema changes on third party applications, the schema integration test system could run critical [data consumer vignettes](https://github.com/HumanCellAtlas/data-consumer-vignettes).
+
+### Pros:
+
+* Bundles only have to be uploaded once
```
tburdett:
```suggestion
* Submissions only have to be ingested once
```

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,80 @@
...
+* Bundles can be uploaded directly to the production environment
```
tburdett:
```suggestion
* Bundles can be pushed directly to the DSS in the production environment
```

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,80 @@
...
+In this design we will use the following terms:
+ 
+ * "schema integration test system" - an automated CI/CD pipeline used to check that downstream systems can handle bundles with a new schema
+ * "new schema label" - anything that indicates that the schema has changes that have not yet been tested against downstream systems: potentially a JSON field, git tag, git branch, or comparison to tested version list
+
+
+Design
+
+1. Bundles with new metadata schema changes are uploaded to the production environment with a new schema label
+1. Observing the new schema label, the DSS stores the new bundles and makes them available with the `GET /v1/bundles` endpoint with the bundle `uuid` or `(uuid, version)`, but does not index nor does it release subscription notifications for them.
```
tburdett:
```suggestion
1. Observing the new schema label, the DSS stores the new data and makes it available with the `GET /v1/bundles` endpoint with the bundle `uuid` or `(uuid, version)`, but does not index nor does it release subscription notifications for these bundles.
```

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,80 @@
...
+1. Via a schema integration test system, a sample of the uploaded bundles are copied from the production DSS to the integration DSS testing environment. Bundles in the sample are selected for unique experimental graph shapes and fields that may have changed.
```
tburdett: One _bundle_ from each _submission_ would probably be sufficient for testing purposes here. It's possible (but very unlikely) that one submission contains multiple graph shapes and/or exercises different subsets of schema. It's perhaps more likely that multiple bundles from the same submission might use different combinations of optional fields.

And as earlier, I'd avoid using the phrase "uploaded bundles".

rfcs/text/0000-metadata-integration-test.md
malloryfreeberg: I agree with @simonjupp on this. From my perspective, I see four "data flow" paths.
1. Ingest -> DS (submit primary data)
1. DS -> Browser (display primary data)
1. DS -> Pipelines -> Ingest -> DS (analyze primary data and submit secondary data)
1. DS -> Browser (display secondary data)

(1) should happen even if (2) - (4) fail. (2) should happen even if (3) and (4) fail (or aren't supposed to happen). (3) should happen even if (4) fails. There might be other "ifs" I'm missing, but the general idea is that failing downstream paths shouldn't prevent upstream paths from happening. I think some amount of retries/replays are necessary as we are always going to be playing "catch up" with new data types being submitted.

tburdett: To echo @malloryfreeberg's point - I'd question whether we really want to prevent the release of raw or analysis data to tertiary portals ("red boxes") because e.g. data cannot be correctly displayed in the browser. I appreciate the technical reasons as to why this might be difficult, so it's sensible to be pragmatic here, but the limitations of this design at least need to be called out.

tburdett (dismissed):
This is a really positive step forward, I'm very supportive of this PR. I've added a few general comments here (and a couple of suggestions) over the language as we discussed at the F2F. I'm a little unclear on the labelling mechanism and I think there's a challenge about the release of data, so I've requested changes, but most overlap with @simonjupp's requested changes and the rest of mine are very minor. Good stuff!

rfcs/text/0000-metadata-integration-test.md
malloryfreeberg: To provide some empirical evidence, 4 of the 16 CBeta projects correctly contained more than 1 unique graph shape. These were due to 2 projects that included both 10x and SS2 data and 2 projects where different samples underwent different numbers of protocols. So, maybe not *very* unlikely. Testing one example of each graph shape in the project would definitely be preferable, although I appreciate that simply finding all unique graph shapes is not a trivial problem. @hewgreen and I have some code that finds unique graph shapes based on a calculating a few graph features. A similar approach could work here.

rfcs/text/0000-metadata-integration-test.md
mweiden: I appreciate the suggestions. Let me try to incorporate them point by point.

> bundles are uploaded to the DSS but not the DCP (they are generated by ingest from _submissions_)

I've removed the word `(bundles)`

> runtime failures aren't exactly unpredictable

My intent here was the was to describe the system as it is now, rather than describing what it could be. Further, I did not take the rewrite wholesale as I don't believe that it addresses the core concern of the RFC: under any of the proposed solutions to metadata schema evolution (schema contract assertions, metadata migrations, metadata "APIs"), we'll need an integration test its assumptions and to gauge impact on critical data consumers.

> what does "the code" refer to

Good catch, replacing this with `metadata schema changes`.

> why do data wranglers and data consumers need to react (wranglers have successfully handed off new data, consumers will get an error and therefore just fail to incorporate the newest data)?

Under the current system:

1. Data wranglers will have to check the progress of an (frequently unreliable) upload, re-upload data, or work with the dev team to get the upload through.
1. I'm not sure that I understand your point about data consumers. Failing to incorporate the newest data and then having to scramble to adapt before the DSS retry window elapses in order to get the new data is the reactive fix I'm speaking of. We may not choose to fix every data consumer before releasing data with a new schema, but at least we'd _know_ that we were about to introduce a break.

rfcs/text/0000-metadata-integration-test.md
mweiden: I think this process is best left up to ingestion/wranglers to decide. I know @daniwelter and @malloryfreeberg had some ideas on how this would work best. What is important to this design is that it is easy to identify from the metadata itself.

From my naïve perspective, it could be as simple as an addition to the path string on the `describedBy` URL. The DSS could then inspect all the submittable types in a bundle and filter accordingly.

rfcs/text/0000-metadata-integration-test.md
mweiden: The important observation here is that we have some strategies and tooling for sampling bundles. We can probably allow developers to gather this feedback and select the best strategy outside of this discussion.

rfcs/text/0000-metadata-integration-test.md
mweiden: @tburdett @diekhans @simonjupp Per our discussion at the F2F, I've added the following sentence/design element. I believe it addresses this concern.

> The DSS subscription API should enable clients to opt into receiving events with unsupported metadata schema versions. Notably, this would enable the data browser to display data with unsupported metadata schema versions even if analysis cannot yet process it.

rfcs/text/0000-metadata-integration-test.md
```diff
@@ -0,0 +1,81 @@
...
+*Integration test data flow: test path is shown in red, production release in blue, integration tooling in dashed lines; edges are labeled with corresponding step numbers in the design above*
+
+1. The DSS subscription API should enable clients to opt into receiving events with unsupported metadata schema versions. Notably, this would enable the data browser to display data with unsupported metadata schema versions even if analysis cannot yet process it.
+1. To estimate the impact of schema changes on third party applications, the schema integration test system could run critical [data consumer vignettes](https://github.com/HumanCellAtlas/data-consumer-vignettes).
+
+### Pros:
+
+* Data files only have to be uploaded once
+* Data files can be uploaded directly to the production environment
+* Tests exactly what data downstream systems will see; does not require complicated fake metadata generation code in tests
```
daniwelter: I'm concerned that using real data in the integration test may not fully test new schema changes. I know programmatically generated metadata isn't ideal but it allows us to have data for all metadata fields whereas the majority of real datasets only supply metadata for a subset of the available metadata fields. It might therefore be possible that a metadata field change isn't actively used for weeks or even months if the change is made following early discussions between wranglers and submitters, but the actual data submission is held up for any of a number of good reasons.

mweiden: Good point. I think it would be totally reasonable for the implementer of the test to add generation or templating code to more fully test the schema changes. The point here was that we could get value/coverage from the test even without such a system.

No integration test will be perfect here. If we can get the 80 from the 80/20 split here, we'd be ahead.

Let's put a pin in this discussion. The implementer should integrate the sampling strategy feedback from you @tburdett and @simonjupp and then figure out how best to go forward. Importantly, we know that there are reasonable ways to approach this problem.

simonjupp (approved):

rfcs/text/0000-metadata-integration-test.md
tburdett: @daniwelter @malloryfreeberg can I ping you for a comment on this? My concern is that this places yet another manual process into the list of things wranglers have to do in order to workaround a technical issue.

rfcs/text/0000-metadata-integration-test.md
tburdett: > Currently, breakages to DCP systems caused by metadata schema changes are detected at runtime upon data upload

My previous comment was based on a broader interpretation (see "submissions making use of new experimental graph shapes may also trigger a breakage without any schema changes") of this RFC than I think perhaps you intend. If this is out of scope, this text is fine. Maybe update the title and the summary to reflect the narrower scope though?

> react to unpredictable runtime failures

I still don't like use of "unpredictable" here. Failures are very predictable - wranglers are investing a large chunk of time in researching which fields components are using to warn about upcoming changes, and the metadata change SOP builds time in for the various changes to allow for developer review and comment. Clearly this process isn't working, so I support better testing, but describing failures as unpredictable is factually incorrect IMO

> Data wranglers will have to check the progress of an (frequently unreliable) upload, re-upload data, or work with the dev team to get the upload through.

I'm not seeing that uploads are frequently unreliable due to metadata schema changes - aren't the vast majority of new schemas successfully integrated into ingest and the DSS? I don't have precise stats on where failures occur but I'm sure they're easy to obtain... there's a large amount of reingestion happening currently but my understanding is that this is due to scaling and representational problems, not unsuccessful upload. Is that not correct?

> data consumers. Failing to incorporate the newest data and then having to scramble to adapt before the DSS retry window elapses in order to get the new data

OK, this makes sense - we're using a different definition of data consumer. I'd consider a data consumer as an end user (possibly an application) but you're thinking specifically of developers of data consuming applications. Maybe state this explicitly?

mweiden: OK, I see. I've made the following changes, which, I hope, address your concerns.

1. Rephrased the first sentence
1. Retitled the RFC "Integration testing for metadata schema changes" - while some systems, like Azul, do rely on the graph structure, I don't think we can reasonably require testing for every new structure, given that almost every experimental structure is unique.
1. Removed the word `unpredictable`
1. Removed the words `(frequently unreliable)`
1. Changed `data consumers` to `Data Storage Service (DSS) API clients`

rfcs/text/0000-metadata-integration-test.md
mweiden: Re. adapting to metadata schema changes as a technical issue: thus far, all of the strategies to protect data consumers (app developers) from changes in the metadata schema we've discussed rest on assumptions or human configuration that only a test can verify:

* Core vs. non-core schema - How do we know that core DCP components or important external data consumers (e.g. Bioconductor) do not rely on non-core schema? They do now.
* Uploader-determined major versioning - even in minor versions adding a field or changing controlled vocabulary from a field could result in broken expectations from data query results. Here's an over-simplified example: say we have all human data in the system, we start ingesting mouse, and add a "species" field. Subscribers that were previously expecting just human data would now have mouse data too (query false positive). Ultimately, metadata major/minor versioning is a human judgement which needs to be tested.
* Likely-used field blacklists/warnlists - How do we know the list is an up to date with downstream systems?
* Assertions on the schema - Client-side assertions on the schema seem like an unattractive option: all consumers (potentially third-party) of the raw metadata would have to store failed events/ids in a dead-letter queue or log for retry after the code is patched. This is not user-friendly.

Given that none of these strategies is a sure-fire way to protect consumers of the metadata from having their expectations broken, having a test isn't a workaround to a technical issue. It's verification in the face of uncertainty.

justincc (approved):

mweiden:
Approved in the 2019-03-08 Architecture Team Meeting.

PR #68, opened 2019-02-14
RFC for DSS extensible external file support

mikebaumann:
RFC for enabling access to files stored externally to the DSS.

Currently open to review/feedback by the UCSC DSS team.

Addresses [HumanCellAtlas/data-store#912]


kozbo:
This PR will not be used/implemented for many months. It was suggested that we close it with an rfc-withdrawn label

PR #77, opened 2019-04-23
RFC Simple Metadata Updates

lauraclarke:
RFC describing the proposed solution to provide simple metadata updates and additions in the DCP. This represents the first step on the road to full Addition, Update, Delete and Retract  (AUDR) functionality in the DCP

Note - note sure of the naming scheme for images so happy to change them as appropriate

~~**May 7th 2019** Last call for community review~~
~~**May 17th 2019** Last call for oversight review~~


rfcs/text/0000-simple-updates.md
```diff
@@ -0,0 +1,159 @@
...
+
+As a data contributor, I would like to correct an error I have made in my submission to ensure that the correct metadata is presented to consumers of my data.
+
+As a data wrangler, I would like to update organ system metadata to reflect an improved ontology annotation for this particular organ.  
+
+Read more on the general user experience background for this work: [User Experience Problem Space - Top task “Update my project”](https://docs.google.com/document/d/1FP6kJa3f9NkOHTLqmkNYztls7soZLgeGefxdgR81ONE/edit)
+
+## Scientific "guardrails"
+
+The proposed design has consequences for end users where different types of updates will have different impact on how data is visible in the DCP. Updates and additions which are in scope for this solution will have provanance trails which could be displayed to users but other updates such as those which change the experimental design representation or require reanalysis will still use the old update process and won't have any provanance reflected inside the DCP.
```
kislyuk: s/provanance/provenance/

rfcs/text/0000-simple-updates.md
```diff
@@ -0,0 +1,159 @@
...
+
+* **Step 5 - Ingest shows the diff to the wrangler for approval**
+
+The wrangler will be shown the difference between their updates and the existing submission in the ingest UI. They will also be shown any validation errors. If there are no validation errors then they will have the option either to approve the update, in which case ingest will start submitting it to the datastore, or cancel it, in which case the update will be removed from ingest and they can return to step 3. If there are validation errors then the wrangler can only cancel the update and submit a new update spreadsheet.
+
+* **Step 6 - Ingest uploads new versions of metadata files and bundles to the datastore**
+
+For this solution, ingest will calculate the bundle updates and submit these to the datastore. This represents a risk to the scalability of this solution. If performance problems emerge we will need to consider solutions in a future RFC.
```
kislyuk, line 94: Can you elaborate on where the risk comes from?

rfcs/text/0000-simple-updates.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Detailed Design
+
+### User Interactions
+
+These are the steps that a wrangler will follow to perform a primary metadata update.
+
+![](../images/0000-simple-updates-ingest-flow.png)
+
+* **Step 1 - Wrangler finds submission**
```
kislyuk, line 69: Steps 1 through 6 here are being described in terms of wranglers interacting with the Ingest Service front-end UI. However, DCP services are expected to expose publicly documented APIs, for purposes of portability and automation.

Can this section be updated to define the actions in terms of the Ingest Service API? I think it's fine to leave the wrangler UI interactions in here, if it's made clear that they build directly on top of specifically described corresponding actions in the Ingest API.

rfcs/text/0000-simple-updates.md
```diff
@@ -0,0 +1,159 @@
...
+
+For this solution, ingest will calculate the bundle updates and submit these to the datastore. This represents a risk to the scalability of this solution. If performance problems emerge we will need to consider solutions in a future RFC.
+
+### DCP Flow of Updates
+
+![](../images/0000-simple-updates-dcp-flow.png)
+
+Once the bundles are updated (1 & 2), the datastore will send update notifications to all listeners (3). The query service, data browser and matrix service **MUST** all process these notifications to provide the updated metadata to consumers as required. The data processing pipeline execution service **MUST NOT** submit new or updated analyses to ingest for assays that have previously had analysis submitted. This solution requires the data processing pipeline execution service to capture and check if notification has already triggered a pipeline run and stop the run if it has.
+
+For simple updates, ingest will update secondary bundles with primary data (copy forward) at the same time that it updates the primary bundles. It does not rely on a signal from the analysis component. This should change in future iterations of the update process.
```
kislyuk, line 109: Can you elaborate on what signal could/would be required from the analysis component in the future?

rfcs/text/0000-simple-updates.md
```diff
@@ -0,0 +1,159 @@
...
+
+* **Step 6 - Ingest uploads new versions of metadata files and bundles to the datastore**
+
+For this solution, ingest will calculate the bundle updates and submit these to the datastore. This represents a risk to the scalability of this solution. If performance problems emerge we will need to consider solutions in a future RFC.
+
+### DCP Flow of Updates
+
+![](../images/0000-simple-updates-dcp-flow.png)
+
+Once the bundles are updated (1 & 2), the datastore will send update notifications to all listeners (3). The query service, data browser and matrix service **MUST** all process these notifications to provide the updated metadata to consumers as required. The data processing pipeline execution service **MUST NOT** submit new or updated analyses to ingest for assays that have previously had analysis submitted. This solution requires the data processing pipeline execution service to capture and check if notification has already triggered a pipeline run and stop the run if it has.
```
kislyuk, line 107: It’s my understanding that analysis pipelines trigger on bundles according to criteria the pipeline developer defines. The pipeline "adapter" marshals data files and metadata fields into the analysis environment, explicitly enumerates them as analysis inputs, and saves a record of this process in the "inputs" component of the output bundle analysis metadata JSON. The same marshaling and explicit record is also made for the analysis outputs. The analysis runs, and the results are submitted to the Ingest Service API to create the output bundle. This process is a key element of our analysis reproducibility strategy.

With this in mind, I suggest that the process be amended as follows, to more rigorously define the practice of event handling by analysis for primary bundle updates process.

Upon receiving an update notification, the Data Processing Service:

- checks if an analysis output bundle already exists in the data store for this input bundle and analysis pipeline version
- if not, runs the pipeline
- if yes, computes the input fields for the pipeline WDL and diffs them against the output bundle’s recorded pipeline inputs
- if the diff is non-empty, runs the pipeline; otherwise, ignores the notification.

To make the implication of this process explicit, any bundle metadata updates that do not affect analysis inputs will not trigger new analysis runs.

I think it's also worth spelling out what happens if a new analysis bundle *is* submitted in this case. A simple solution would be to retain the output bundle UUID, but completely replace its contents with new outputs.

I understand that this RFC is shooting for a minimalistic solution for simple metadata updates, and I support that. I also think we can make the temporary simplifying assumption that these simple metadata updates won't change pipeline output for the limited set of pipelines that we have now. But if we make this assumption here, then we should explicitly acknowledge this as a limitation that should be resolved in the near future, and draft a follow-on RFC for that. (Happy to help do so)

kislyuk (commented):
One more point, I really like this RFC overall and I think it's a good minimalistic way to describe the initial update functionality. I definitely agree with the idea of splitting the AUDR/data model/data lifecycle discussion into a series of small, easy to handle RFCs.

rfcs/text/0000-simple-updates.md
justincc: 1. Certain metadata changes, such as changes to project metadata, will require updates to all bundles in that project. 
2. Some projects have a large number of bundles (e.g. Tabula Muris at ~100K).
3. We already know from [a previous situation where we had to identify all Tabula Muris bundles for recreation then tombstoneing](https://app.zenhub.com/workspaces/dcp-5ac7bcf9465cb172b77760d9/issues/humancellatlas/ingest-central/343) that this can take a considerable time, morethan a couple of hours but less than a day (uncertain because this wasn't a performance measurement).
4. So it's possible that metadata updates which touch all or a very large percentage of bundles in a project with many bundles may take us a long time to complete on the ingest side. At the very least, we have to make 100K of update requests and more 100K requests when we allow analysis updates, assuming copy-forward remains. However, the problem above is different - it was about bundle creation using files already in the datastore, not update) and there may be work we can do ingest side to improve performance. Also, what is acceptable performance is yet to be determined.

There was a [ticket](https://github.com/HumanCellAtlas/data-store/issues/1991) open with datastore to investigate an alternative. This was for them to provide an API where ingest could specify a set of file version updates for which all bundles containing the old version should be updated to point to the new version. This would save ingest from having to construct this list out of its own bundle state.

However, this in itself might require significant design and performance assessment. So rather than delay updates further I opted to proceed with doing this task in ingest and seeing what we can achieve in terms of performance. The risk is that we can't achieve good performance or doing so would be prohibitively costly, at which point one option is to revisit the topic of datastore updating bundles if it is much more efficient/realistic for them to do so.

rfcs/text/0000-simple-updates.md
justincc: This RFC is concentrating on the requirements as they exist now, which is to allow wranglers to make updates. There are no users of this functionality in Q2 apart from the ingest UI.

Hence, I think it is better to keep these APIs private (unpublished) until other components need to use them. At that point, we can revise them in the light of our experience before RFCing and publishing them. Increasing out public API surface at this point will make it more difficult to do future revisions without any obvious benefit. It's also the case that these details are not yet pinned down at the low level as ingest design for updates is complex. Fully forming this API to a publishable level will take extra time.

That's my opinion. @lauraclarke, @morrisonnorman please could you comment.

rfcs/text/0000-simple-updates.md
justincc: In the future, when analysis can process updates:

a) If copy-forward is still in place, analysis needs to signal that they decided not to re-run analysis in response to an update, at which point ingest needs to update the primary data in all the analysis bundles.
b) If analysis does run in response to an update, ingest needs to know:
  i) if the analysis is entirely new (not an update to an existing analysis) or
  ii) is the analysis updates an existing analysis.
Ingest cannot make that determination itself.
c) If an analysis is an update, ingest needs to know which files in the new analysis are updates to files in the existing analysis, so it can tell that datastore that these are file updates rather than entirely new files. Simplification with cast iron rules might be possible, e.g. that any file in the new analysis with the same path is an update to the file with the same path in the previous analysis, that any file in the old analysis but not in the new should be dropped. If we can't adopt these cast iron rules then the submitter (analysis) will need to specify - ingest cannot make those determinations.

rfcs/text/0000-simple-updates.md
justincc: Yes, I believe planners deliberately restricted updates scope to avoid having to decide on these questions right now, which were generating a lot of discussion without obvious conclusions. @lauraclarke, @morrisonnorman could you make sure this limitation is explicit?

Regarding your comment:

> I think it's also worth spelling out what happens if a new analysis bundle is submitted in this case. A simple solution would be to retain the output bundle UUID, but completely replace its contents with new outputs.

Yes, this was one solution I was thinking about before the Q2 scope restriction. It may avoid the need to make decisions about analysis file updates discussed above. I guess it depends on how this would affect consumers.

rfcs/text/0000-simple-updates.md
kislyuk: The API could be marked as developmental/provisional, but the API needs to be there ("public" in the sense of available for applications other than the first-party UI to call, not "open to all public at large"). The users of this functionality are/will be the wranglers, who can't develop automation for their work without a documented API. I think at this stage it's very reasonable for that API to change quickly in breaking ways.

rfcs/text/0000-simple-updates.md
ambrosejcarr: > Yes, this was one solution I was thinking about before the Q2 scope restriction

This discussion seems reasonable given the scope restriction. I want to flag, for the future, that it is likely that analysis pipelines will consume metadata, such as batch, replicate, or patient information, to carry out normalization. In these cases, changes to certain metadata fields **MUST** provoke reanalysis.

lauraclarke: @ambrosejcarr We are in full agreement that this is functionality which will be required. We decided that figuring out the logic for reanalysis vs analysing with a new analysis pipeline (and if the old analysis results do or don't need to be reacted) was sufficiently complex that we didn't want it to delay the simple updates.  We must make sure that conversation starts and we can get an RFC about the plan for solving that problem.

diekhans:
Can the source files for the images be added to the tree?  It is problematic not to have the source so that images can be changed in the future.

diekhans:
The RFC seems to not define how information is recorded as to _**why**_ a change was made.   This knowledge is key to understand the provenance, as opposed to just tracking it.  In the long run, we probably want to make this information public and browseable along with the metadata changes.  In the short run, just capturing it in a way that can be mapped back to revisions seems minimal.

justincc:
> The RFC seems to not define how information is recorded as to _**why**_ a change was made. This knowledge is key to understand the provenance, as opposed to just tracking it. In the long run, we probably want to make this information public and browseable along with the metadata changes. In the short run, just capturing it in a way that can be mapped back to revisions seems minimal.

Without commenting on whether we want to do this right now (I'll leave that to @lauraclarke and @morrisonnorman), capturing the why will require UX consideration (where does the wrangler state this (spreadsheet, UI?)) and some internal consideration (where do we keep it?). It will almost certainly mean implementation for the RFC will not fit in Q2.

diekhans:
Recording why I wasn't intending why to be anything more complex
than English language comments that get associated with the
change.  Not losing human knowledge when there has been full
turnover of staff.

This could be as simple as a text file stored in some git repo,
as longer as there is some way to map it back to the change in
the future.  When/if we have a more formal way of tracking this,
it can be reformatted for future generations.

This is more SOP than software engineering.  In the future, it
would be good to a more formal mechanism to make it easier to
record and display, but for now, it is like a change log.

rfcs/text/0000-simple-updates.md
lauraclarke: @kislyuk I agree with Justin at this early point in the design this introduces some risk especially as during Q2 there are no expected programmatic users.

This is because wranglers only interact with the service via the UI. Currently, we only have one user of the programmatic service, the pipelines team and as they aren't expecting to submit updates this quarter.  

This means avoiding specifying the endpoints now mitigating the risk of the design needing to change. If we were immediately expecting programmatic users and for them to comment if the endpoints were reasonable it would be different but that isn't the position we are in.

rfcs/text/0000-simple-updates.md
```diff
@@ -0,0 +1,161 @@
...
+## User Stories
+
+As a data contributor, I would like to add the details of a scientific publication to my submission to the DCP to broaden knowledge of the work.
+
+As a data contributor, I would like to correct an error I have made in my submission to ensure that the correct metadata is presented to consumers of my data.
+
+As a data wrangler, I would like to update organ system metadata to reflect an improved ontology annotation for this particular organ.  
+
+Read more on the general user experience background for this work: [User Experience Problem Space - Top task “Update my project”](https://docs.google.com/document/d/1FP6kJa3f9NkOHTLqmkNYztls7soZLgeGefxdgR81ONE/edit)
+
```
brianraymor, line 53: In the _user experience background_ material reference above, this is commented as MVP:
 
> Edit the data file names for existing files

Could you offer an example -and- is this in scope for this RFC?

rfcs/text/0000-simple-updates.md
```diff
@@ -0,0 +1,161 @@
...
+* The wranglers can access the spreadsheet for a particular submission via the ingest UI. The returned spreadsheet will contain all the relevant information to enable the ingest service to carry out the ingest and will use the schema of the original submission.
+* The wranglers can see what documents in a submission have changed via an ingest UI view. 
+
+
+### Unresolved Questions
+
+This RFC does not represent a complete solution to the Addition and Update functionality that is needed by the DCP. There are a number of unresolved questions and missing functionality that will be left for future RFCs to address.
+
+*   How is the provenance of data and metadata made clear to consumers?
+*   How are data retracted without creating meaningless 404 errors for consumers who had used it? [RFC4: Deletion of Data in the DCP]([https://github.com/HumanCellAtlas/dcp-community/blob/master/rfcs/text/0004-dss-deletion-process.md](https://github.com/HumanCellAtlas/dcp-community/blob/master/rfcs/text/0004-dss-deletion-process.md)) considers some of this challenge.
```
brianraymor: `[RFC4: Deletion of Data in the DCP]([https://github.com/HumanCellAtlas/dcp-community/blob/master/rfcs/text/0004-dss-deletion-process.md](https://github.com/HumanCellAtlas/dcp-community/blob/master/rfcs/text/0004-dss-deletion-process.md))` should be:

`[RFC4: Deletion of Data in the DCP](https://github.com/HumanCellAtlas/dcp-community/blob/master/rfcs/text/0004-dss-deletion-process.md)`

rfcs/text/0000-simple-updates.md
```diff
@@ -0,0 +1,161 @@
...
+2. Updates which require additional or updated experimental data files
+3. Updates which require an analysis pipeline to be run again, either a new pipeline or the same pipeline with different parameters
+4. Migrating metadata to a new schema where this is necessary before a metadata update can occur. This is being considered independently of updates.
+5. Removing experimental data files from a submission.
+6. Redacting metadata (in the sense of completely purging it/hard deleting it from the system)
+7. Redacting data (in the sense of completely purging it/hard deleting it from the system)
+8. Updating metadata on unsubmitted datasets.
+
+Implementing all of these features will take the DCP time and we feel that the proposed design represents a step forward in the update and additional process and is valuable even if not complete. One important issue this design allows us to address which the old update process does not is it will give us an auditable trail of edits to bundles which can be displayed to users, improving their experience of and trust in the platform.
```
brianraymor: *Editorial*: Implementing all of these features will take ~the DCP~ time ~and~, but ~we feel that~ the proposed design represents a step forward in the update and addition~al~ process ~and~, which is valuable even if ~not~ incomplete. ~One important issue this design allows us to address which the old update process does not is it will give us an auditable trail of edits to bundles which can be displayed to users~, Unlike the current update process, this design offers an auditable trail of edits to bundles which can be presented to users. improving their experience of and trust in the platform.

rfcs/text/0000-simple-updates.md
```diff
@@ -0,0 +1,161 @@
...
+
+As a data contributor, I would like to correct an error I have made in my submission to ensure that the correct metadata is presented to consumers of my data.
+
+As a data wrangler, I would like to update organ system metadata to reflect an improved ontology annotation for this particular organ.  
+
+Read more on the general user experience background for this work: [User Experience Problem Space - Top task “Update my project”](https://docs.google.com/document/d/1FP6kJa3f9NkOHTLqmkNYztls7soZLgeGefxdgR81ONE/edit)
+
+## Scientific "guardrails"
+
+The proposed design has consequences for end users where different types of updates will have different impact on how data is visible in the DCP. Updates and additions which are in scope for this solution will have provenance trails which could be displayed to users but other updates such as those which change the experimental design representation or require reanalysis will still use the old update process and won't have any provanance reflected inside the DCP.
```
brianraymor: Spelling. Second instance of *provanance* -> provenance - also, *auditable trail* is used below so you may wish to choose a common phrase. But foolish consistency and hobgoblins ... And that's a really long sentence that I'd break up for easier reading.

mckinsel (approved):
Talked through the matrix service bits with the team, looks good to us.

lauraclarke:
> Recording why I wasn't intending why to be anything more complex than English language comments that get associated with the change. Not losing human knowledge when there has been full turnover of staff. This could be as simple as a text file stored in some git repo, as longer as there is some way to map it back to the change in the future. When/if we have a more formal way of tracking this, it can be reformatted for future generations. This is more SOP than software engineering. In the future, it would be good to a more formal mechanism to make it easier to record and display, but for now, it is like a change log. Justin Clark-Casey <notifications@github.com> writes:
> […](#)
> > The RFC seems to not define how information is recorded as to _**why**_ a change was made. This knowledge is key to understand the provenance, as opposed to just tracking it. In the long run, we probably want to make this information public and browseable along with the metadata changes. In the short run, just capturing it in a way that can be mapped back to revisions seems minimal. Without commenting on whether we want to do this right now (I'll leave that to @lauraclarke and @morrisonnorman), capturing the why will require UX consideration (where does the wrangler state this (spreadsheet, UI?)) and some internal consideration (where do we keep it?). It will almost certainly mean implementation for the RFC will not fit in Q2. -- You are receiving this because you commented. Reply to this email directly or view it on GitHub: [#77 (comment)](https://github.com/HumanCellAtlas/dcp-community/pull/77#issuecomment-488221907) The RFC seems to not define how information is recorded as to why a change was made. This knowledge is key to understand the provenance, as opposed to just tracking it. In the long run, we probably want to make this information public and browseable along with the metadata changes. In the short run, just capturing it in a way that can be mapped back to revisions seems minimal. Without commenting on whether we want to do this right now (I'll leave that to @lauraclarke and @morrisonnorman), capturing the why will require UX consideration (where does the wrangler state this (spreadsheet, UI?)) and some internal consideration (where do we keep it?). It will almost certainly mean implementation for the RFC will not fit in Q2. — You are receiving this because you commented. Reply to this email directly, view it on GitHub, or mute the thread.*

@diekhans while I fully understand the desire here I am trying to figure out a way to do it which will persist when contributors themselves become able to directly edit without wrangler intervention and I worry that any system we put in place now will result in false expectations about reasoning being recorded and useful in the long term. I certainly can't think of any archive which stores a decision record for updates. It is a lovely idea and I am just struggling to see the uptake to make the additional effort valuable

rfcs/text/0000-simple-updates.md
lauraclarke: @gabsie is the owner of that document and I am uncertain where the requirement for file name changes came from, hopefully, she can comment.

As far as MVP goes, these other documents made statements about what might be considered MVP but this RFC reflects the scope for the current work which is restricted based on capacity vs other documents.

rfcs/text/0000-simple-updates.md
lauraclarke: This should be fixed

rfcs/text/0000-simple-updates.md
lauraclarke: Thanks for the suggestion, I hope I transferred it faithfully

rfcs/text/0000-simple-updates.md
lauraclarke: Should be fixed and I made the audit trail consistent with provenance, happy for you to make further suggestions to make the text more readable

rfcs/text/0000-simple-updates.md
gabsie: @brianraymor @lauraclarke @justincc 
Hey. So I put this in my initial problem space doc, since I wanted to address the scenario of making a human error in the names of things. This is just one example: putting the wrong filename in the sheet with filenames and needing to correct it.

If this is not covered by the current simple update scenario, then I am OK to edit things in my doc re: priority. I can definitely see how initial submitters or even data wranglers can make human errors when pasting/spelling/putting the wrong filename across the sample/etc. I have been a witness of such in a couple of discussions between wrangler and submitter.

Let me know what to do next, my document is quite high level and deals with general user needs, which are very true, but which might not necessarily be addressed in this first phase.

brianraymor: No further action is required for the RFC from my perspective since Laura confirmed that the scope is restricted based on the RFC and not the reference material. I simply wanted to confirm.

rfcs/text/0000-simple-updates.md
kislyuk: That sounds good, thanks for clarifying. I think we may need to have a joint discussion about scalability issues here later, including the need to remove the "one bundle per cell" mode of operation in SmartSeq2 as it seems to be detrimental to scalability.

rfcs/text/0000-simple-updates.md
kislyuk: In my opinion it is important to always define an API first, and then define a UI (and possibly some UI-specific "value added" service functionality) on top of it. This allows one not to get into uncomfortable situations where the reason no automation exists is that no tools were available for automation, or where an API has to be retrofitted on top of a system that organically evolved to fuse API and UI design decisions together. I think it's also important that we get away from having the spreadsheet-centric paradigm of data handling be the lowest tier of the ingest service data interchange model. Spreadsheets are a fine way to work with metadata, but the underlying system and API should be able to accept the constituent metadata in its raw form as arrays or associative arrays of strings.

The ingest service does have a commitment to define an API as part of its charter, and I think an API sketch should be the starting point of any RFC of this type. We should get into the habit of centering our discussions around that.

With all that said, I understand the time and resource constraints that prompted this RFC's scope restrictions. To clarify the intent, I propose adding a new section to this RFC, stating that API design for this feature is postponed but forthcoming. Here is my proposed wording - please feel free to change it as you see fit:

--------------

## Detailed Design
### Metadata Updates API

The Ingest Service API provides a key capability for data wranglers to provide wrangling portals (interfaces) appropriate for their specific user or data modality needs, as well as for automation of bulk/batch data submission, data QC and status monitoring. The Ingest Service API also provides programmatic access for the analysis pipeline execution manager (however, metadata updates from analysis pipelines are beyond the scope of this RFC).

The Ingest Service API and its attendant developer resources ([documentation](https://github.com/HumanCellAtlas/ingest-core) and [API client](https://github.com/HumanCellAtlas/ingest-client)) will be updated to accommodate metadata update functionality as described in this RFC. The design of these API updates is beyond the scope of this RFC, and will be addressed in a future RFC.

--------------

cc @claymfischer @cjvillar

kislyuk (approved):
Approved with the understanding that the RFC will be updated to reflect API availability questions

justincc:
> Approved with the understanding that the RFC will be updated to reflect API availability questions

Added in text following your proposal. Thanks.

theathorn (approved):

hannes-ucsc (approved):
This RFC mentions _updating files and bundles in the data store_. Bundles and files in the data store are identified by a UUID and a version. I assume _updating files and bundles_ implies that the UUID of of the the updated copy of a file/bundle is the same as that of the original, and that only their version differs. If that's correct,  I approve this RFC.

sampierson (approved):

rfcs/text/0000-simple-updates.md
samanehsan: Just to clarify `The data processing pipeline execution service **MUST NOT** submit new or updated analyses to ingest for assays that have previously had analysis submitted.` If an update comes in and only the pipeline version has changed, is the expectation here that we would not re-analyze the data with the updated pipeline until a mechanism is in place to handle re-analysis?

samanehsan (approved):
This looks good! I just had one question here: https://github.com/HumanCellAtlas/dcp-community/pull/77#discussion_r285257291

justincc:
> This RFC mentions _updating files and bundles in the data store_. Bundles and files in the data store are identified by a UUID and a version. I assume _updating files and bundles_ implies that the UUID of of the the updated copy of a file/bundle is the same as that of the original, and that only their version differs. If that's correct, I approve this RFC.

Sorry missed your comment at the time. Yes, this is correct - updates will have the same UUID.

rfcs/text/0000-simple-updates.md
justincc: Yep

PR #85, opened 2019-07-15
RFC: DSS metrics

kislyuk:
Proposes key metrics to track DSS performance, and a process for monitoring them.

Last call for community review: August 23

### DIFF
--- a/rfcs/text/0000-dss-metrics.md
+++ b/rfcs/text/0000-dss-metrics.md
@@ -21,7 +21,7 @@ which the utility of the Data Store can be evaluated, and by which informed data
 
 ## Motivation
 
-Beyond latency and error rates, the DSS lacks a clear focus on quantifiable assessment of usability. We would like to
+Today, the DSS only tracks latency and error rates, which alone, does not provide a full picture of its utility.
 quantify higher level performance metrics that reflect the Data Store's ability to make HCA data accessible, computable,
 and durable.
 
--- a/rfcs/text/0000-dss-metrics.md
+++ b/rfcs/text/0000-dss-metrics.md
@@ -16,11 +16,8 @@ which the utility of the Data Store can be evaluated, and by which informed data
 * [Andrey Kislyuk](mailto:akislyuk@chanzuckerberg.com)
 
 ## Shepherd
-***Leave this blank.** This role is assigned by DCP PM to guide the **Author(s)** through the RFC process.*
 
-*Recommended format for Shepherds:*
-
- `[Name](mailto:username@example.com)`
+* [Andrey Kislyuk](mailto:akislyuk@chanzuckerberg.com)
 
 ## Motivation
 
--- a/rfcs/text/0000-dss-metrics.md
+++ b/rfcs/text/0000-dss-metrics.md
@@ -9,7 +9,8 @@
 ## Summary
 
 This RFC proposes a set of key performance metrics to be introduced and tracked as a summary measure of performance by
-which the utility of the Data Store can be evaluated, and by which informed data-driven improvement and/or regression decisions may be made.
+which the utility of the Data Store can be evaluated, and by which informed data-driven improvement and/or regression
+decisions may be made.
 
 ## Author(s)
 
@@ -24,6 +25,14 @@ which the utility of the Data Store can be evaluated, and by which informed data
 Today, the DSS only tracks latency and error rates, which alone, does not provide a full picture of its utility.
 quantify higher level performance metrics that reflect the Data Store's ability to make HCA data accessible, computable,
 and durable.
+
+This RFC introduces a first iteration of DSS-specific key performance indicators (KPIs) that can be used to make DSS
+performance more measurable, and changes more observable and actionable. The suggested set of metrics is intended to be
+mutable; the service team should be empowered to adjust and expand the set as long as observability is not negatively
+impacted.
+
+The RFC is also intended as a model for establishing service-specific key performance indicators (KPIs) and preparing
+them for roll-up into DCP-wide KPIs.
 
 ### User Stories
 
--- a/rfcs/text/0000-dss-metrics.md
+++ b/rfcs/text/0000-dss-metrics.md
@@ -70,7 +70,7 @@ performance metrics in 3 categories:
 
   Units: US$/GB
 
-Measure the above metrics on both AWS and GCP, separately and in composite.
+Measure the above metrics on both AWS and GCP, separately and in composite (weighted average).
 
 A blend of these metrics can be thought of as the key performance indicator (KPI) for the Data Store to optimize. More
 metrics can be added or changed in each category over time, provided that observability of the system is not reduced.
--- a/rfcs/text/0000-dss-metrics.md
+++ b/rfcs/text/0000-dss-metrics.md
@@ -9,7 +9,7 @@
 ## Summary
 
 This RFC proposes a set of key performance metrics to be introduced and tracked as a summary measure of performance by
-which the utility of the Data Store can be evaluated, including improvements and regressions.
+which the utility of the Data Store can be evaluated, and by which informed data-driven improvement and/or regression decisions may be made.
 
 ## Author(s)
 
--- a/rfcs/text/0000-dss-metrics.md
+++ b/rfcs/text/0000-dss-metrics.md
@@ -22,13 +22,13 @@ decisions may be made.
 
 ## Motivation
 
-Today, the DSS only tracks latency and error rates, which alone, does not provide a full picture of its utility.
-quantify higher level performance metrics that reflect the Data Store's ability to make HCA data accessible, computable,
-and durable.
+Today, the DSS only tracks latency and error rates, which alone, do not provide a full picture of its utility nor allow
+us to quantify higher level performance metrics that reflect the Data Store's ability to make HCA data accessible,
+computable, and durable.
 
 This RFC introduces a first iteration of DSS-specific key performance indicators (KPIs) that can be used to make DSS
 performance more measurable, and changes more observable and actionable. The suggested set of metrics is intended to be
-mutable; the service team should be empowered to adjust and expand the set as long as observability is not negatively
+mutable; the service team should adjust and expand the set over time while ensuring observability is not negatively
 impacted.
 
 The RFC is also intended as a model for establishing service-specific key performance indicators (KPIs) and preparing
@@ -64,15 +64,15 @@ performance metrics in 3 categories:
   Units: rate (number of API requests raising ultimate error (rolled up with their retries) / total API requests)
 
 - *Costing metrics*: Measure "all-in" cost per GB stored (the total cost of DSS services provided divided by the number
-  of GB stored in each replica). (Other metrics or breakout categories could be needed here, such as breaking out and
+  of GB stored in each replica). Other metrics or breakout categories could be needed here, such as breaking out and
   tracking transfer or subscription notification costs separately, but only if and when it is shown that such costs are
-  a significant component of the overall cost footprint.)
+  a significant component of the overall cost footprint.
 
   Units: US$/GB
 
 Measure the above metrics on both AWS and GCP, separately and in composite (weighted average).
 
 A blend of these metrics can be thought of as the key performance indicator (KPI) for the Data Store to optimize. More
-metrics can be added or changed in each category over time, provided that observability of the system is not reduced.
+metrics shouldbe added or changed in each category over time, provided that observability of the system is not reduced.
 
 ### Unresolved Questions
--- a/rfcs/text/0000-dss-metrics.md
+++ b/rfcs/text/0000-dss-metrics.md
@@ -48,24 +48,31 @@ them for roll-up into DCP-wide KPIs.
 ## Detailed Design
 
 Introduce infrastructure (such as logging, instrumentation, offline log analytics, etc.) to measure the following DSS
-performance metrics:
+performance metrics in 3 categories:
 
-- Measure TTFB (time to first byte) along with any dimensions that affect checkout service behavior (file type, size,
-  etc.), User Agent string, and other relevant request data. Time to first byte is defined here as the time from the
-  first request to access a bundle or file to the time of the first request that returns a pre-signed or direct object
-  storage access URL. (This may require an opaque session identifier to be returned when serving responses with a
-  Retry-After header.)
+- *API performance metrics*: Measure TTFB (time to first byte) along with any dimensions that affect checkout service
+  behavior (file type, size, etc.), User Agent string, and other relevant request data. Time to first byte is defined
+  here as the time from the first request to access a bundle or file to the time of the first request that returns a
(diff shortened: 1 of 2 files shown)


rfcs/text/0000-dss-metrics.md
```diff
@@ -0,0 +1,65 @@
...
+## Author(s)
+
+* [Andrey Kislyuk](mailto:akislyuk@chanzuckerberg.com)
+
+## Shepherd
+***Leave this blank.** This role is assigned by DCP PM to guide the **Author(s)** through the RFC process.*
+
+*Recommended format for Shepherds:*
+
+ `[Name](mailto:username@example.com)`
```
maniarathi: Replace with your name for posterity.

rfcs/text/0000-dss-metrics.md
```diff
@@ -0,0 +1,65 @@
...
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC number and this pull request in the format:*
+
+`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+
+# RFC: Key performance metrics of the DCP Data Store (DSS)
+
+## Summary
+
+This RFC proposes a set of key performance metrics to be introduced and tracked as a summary measure of performance by
+which the utility of the Data Store can be evaluated, including improvements and regressions.
```
maniarathi:
```suggestion
which the utility of the Data Store can be evaluated, and by which informed data-driven improvement and/or regression decisions may be made.
```

rfcs/text/0000-dss-metrics.md
```diff
@@ -0,0 +1,65 @@
...
+## Shepherd
+***Leave this blank.** This role is assigned by DCP PM to guide the **Author(s)** through the RFC process.*
+
+*Recommended format for Shepherds:*
+
+ `[Name](mailto:username@example.com)`
+
+## Motivation
+
+Beyond latency and error rates, the DSS lacks a clear focus on quantifiable assessment of usability. We would like to
```
maniarathi: Maybe simplify the language a bit? 

```suggestion
Today, the DSS only tracks latency and error rates, which alone, does not provide a full picture of its utility.
```

rfcs/text/0000-dss-metrics.md
```diff
@@ -0,0 +1,65 @@
...
+quantify higher level performance metrics that reflect the Data Store's ability to make HCA data accessible, computable,
+and durable.
+
+### User Stories
+
+* As a DSS developer, I would like to know whether the DSS is improving in its ability to serve downstream users and
+  services.
+
+* As a downstream service developer, I would like to suggest a way of measuring DSS performance that reflects my ability
```
maniarathi: I like this user story a lot but I don't see any mention in the detailed design of how the metrics provided allows for flexibility to suggest new metrics.

Perhaps this user story is more intended for another RFC around how to request for new metrics instead of this RFC which enumerates the initial set of metrics.

rfcs/text/0000-dss-metrics.md
maniarathi: I would like a better understanding of what these metrics are (is latency per API call? what are the units of the error rates? etc.) as well as a brief description on how they are tracked today. Maybe this could be added to the Design section.

rfcs/text/0000-dss-metrics.md
```diff
@@ -0,0 +1,65 @@
...
+  to use the DSS to consume HCA data.
+
+* As a DCP funder, I would like to know how the DSS is balancing the goals of economical data storage and data
+  accessibility and computability.
+
+## Detailed Design
+
+Introduce infrastructure (such as logging, instrumentation, offline log analytics, etc.) to measure the following DSS
+performance metrics:
```
maniarathi: I think overall, I don't disagree with any of the listed metrics here but I'd like to see a slightly more built out description and perhaps even more metrics. You could perhaps format them in a table where the columns are: Metric Name, Description, Units. 

Underneath the table, I would enumerate the metrics names again and bullet point under each one, the expected infrastructure changes that need to take place. E.g.

**Unrecoverable Error Rate**
- May require added instrumentation in the HCA CLI

rfcs/text/0000-dss-metrics.md
```diff
@@ -0,0 +1,65 @@
...
+
+- Measure unrecoverable error rate, i.e. the rate of error that caused the client to stop trying and cause a client-side
+  error to be raised. (This may require instrumentation in the HCA CLI.)
+
+- Measure "all-in" cost per GB stored (the total cost of DSS services provided divided by the number of GB stored in
+  each replica). (Other metrics or breakout categories could be needed here, such as breaking out and tracking transfer
+  or subscription notification costs separately, but only if and when it is shown that such costs are a significant
+  component of the overall cost footprint.)
+
+Measure the above metrics on both AWS and GCP, separately and in composite.
```
maniarathi: Maybe elaborate a bit more on what you mean by "in composite"

rfcs/text/0000-dss-metrics.md
```diff
@@ -0,0 +1,65 @@
...
+
+A blend of these metrics can be thought of as the key indicator for the Data Store to optimize.
```
maniarathi: This sentence a bit redundant from the objective which states that the metrics will be used for data-driven improvements so you could remove it.

rfcs/text/0000-dss-metrics.md
```diff
@@ -0,0 +1,65 @@
...
+
+* As a DCP funder, I would like to know how the DSS is balancing the goals of economical data storage and data
+  accessibility and computability.
+
+## Detailed Design
+
+Introduce infrastructure (such as logging, instrumentation, offline log analytics, etc.) to measure the following DSS
+performance metrics:
+
+- Measure TTFB (time to first byte) along with any dimensions that affect checkout service behavior (file type, size,
```
maniarathi: I'd like to see even more metrics around the API to support the first user story; for example, per API: max latency, average latency, time to first byte, some measurement of usage, error rates per some unit of time, error percentages, etc.

rfcs/text/0000-dss-metrics.md
maniarathi: On second thought, you could potentially even bucket them into the three overall categories if it makes sense: 1) API Metrics, 2) Error Metrics, 3) Costing Metrics.

maniarathi (changes requested):

samanehsan:
@kislyuk I noticed the PR description says community review ended on 8/23, but it still has the rfc-community-review flag. Was the review period extended?

rfcs/text/0000-dss-metrics.md
kislyuk: Done.

rfcs/text/0000-dss-metrics.md
kislyuk: Added a paragraph to the motivation section discussing this.

rfcs/text/0000-dss-metrics.md
kislyuk: Regarding the re-formatting: how does the re-formatting improve the RFC? It's hard to work with Markdown tables that contain more than a small snippet of text. While readability of the rendered text may be improved a little bit, readability of the Markdown source is drastically reduced.

Regarding expanding the descriptions: I don't mind adding more detail, but what specifically would you like to see?

I've added units to each listing, and augmented the metrics with a list of categories that you mentioned (thanks!)

Regarding adding more metrics: I suggest not overloading this RFC with too many metrics; this is intended as a starting point for KPI measurement. Teams should be free to add more metrics after the first set is implemented.

rfcs/text/0000-dss-metrics.md
kislyuk: Please see my other comment about not overloading this initial RFC with too many metrics to implement. I think it would be great if the team agrees to add more metrics in team-specific discussions.

rfcs/text/0000-dss-metrics.md
kislyuk: Thanks - Clarified this sentence (I really meant weighted average).

rfcs/text/0000-dss-metrics.md
kislyuk: Expanded this sentence to be less redundant and more to unpack the objective.

kislyuk:
@samanehsan no, I just didn't get around to addressing the comments until now. Issuing a call for oversight review now.

rfcs/text/0000-dss-metrics.md
```diff
@@ -0,0 +1,78 @@
...
+Introduce infrastructure (such as logging, instrumentation, offline log analytics, etc.) to measure the following DSS
+performance metrics in 3 categories:
+
+- *API performance metrics*: Measure TTFB (time to first byte) along with any dimensions that affect checkout service
+  behavior (file type, size, etc.), User Agent string, and other relevant request data. Time to first byte is defined
+  here as the time from the first request to access a bundle or file to the time of the first request that returns a
+  pre-signed or direct object storage access URL. (This may require an opaque session identifier to be returned when
+  serving responses with a Retry-After header.)
+
+  Units: seconds
```
DailyDreaming: Should this be changed to megabyte/second to normalize?  One 22Gb file might be an outlier that would seem to throw this off otherwise.

rfcs/text/0000-dss-metrics.md
```diff
@@ -0,0 +1,78 @@
...
+  a significant component of the overall cost footprint.)
+
+  Units: US$/GB
+
+Measure the above metrics on both AWS and GCP, separately and in composite (weighted average).
+
+A blend of these metrics can be thought of as the key performance indicator (KPI) for the Data Store to optimize. More
+metrics can be added or changed in each category over time, provided that observability of the system is not reduced.
+
+### Unresolved Questions
```
DailyDreaming: Should this have more implementation details?  For example, do we graph the data in Google Analytics, AWS Custom Metrics, Grafana... ?

rfcs/text/0000-dss-metrics.md
kislyuk: That doesn't make sense to me; the TTFB metric is intended to measure startup latency (time to first byte), not rolling latency or bandwidth.

rfcs/text/0000-dss-metrics.md
kislyuk: The intent of this RFC is to define KPIs to be measured and exposed. I think prescribing the specific way in which to expose the KPIs may be overly constraining and outside the scope of this RFC. But if you have suggested edits that you would like to use to nail down uncontroversial details, please go ahead and add them for consideration.

rfcs/text/0000-dss-metrics.md
kislyuk: To elaborate: I think it is reasonable to histogram TTFB by file size for information purposes, but the KPI is the average TTFB. When reduced to one number, TTFB (or 95th/99th percentile TTFB) getting skewed by outliers is intentional - the outliers are what matters here.

DailyDreaming: That makes sense, thanks for taking the time to explain.

rfcs/text/0000-dss-metrics.md
DailyDreaming: Alright.

rfcs/text/0000-dss-metrics.md
maniarathi: LGTM

rfcs/text/0000-dss-metrics.md
maniarathi: Makes sense. I was thinking that there would be even more metrics here but with your statement that this RFC not be overloaded the re-format suggestion doesn't make sense anymore.

rfcs/text/0000-dss-metrics.md
maniarathi: LGTM

rfcs/text/0000-dss-metrics.md
```diff
@@ -0,0 +1,78 @@
...
+* [Andrey Kislyuk](mailto:akislyuk@chanzuckerberg.com)
+
+## Shepherd
+
+* [Andrey Kislyuk](mailto:akislyuk@chanzuckerberg.com)
+
+## Motivation
+
+Today, the DSS only tracks latency and error rates, which alone, does not provide a full picture of its utility.
+quantify higher level performance metrics that reflect the Data Store's ability to make HCA data accessible, computable,
```
maniarathi: Incomplete sentence?

rfcs/text/0000-dss-metrics.md
```diff
@@ -0,0 +1,78 @@
...
+and durable.
+
+This RFC introduces a first iteration of DSS-specific key performance indicators (KPIs) that can be used to make DSS
+performance more measurable, and changes more observable and actionable. The suggested set of metrics is intended to be
+mutable; the service team should be empowered to adjust and expand the set as long as observability is not negatively
```
maniarathi: Instead of stating "be empowered" which almost has a feel of giving permission, what about "the service team should adjust and expand the set over time while ensuring observability is not negatively impacted."?

rfcs/text/0000-dss-metrics.md
```diff
@@ -0,0 +1,78 @@
...
+  of GB stored in each replica). (Other metrics or breakout categories could be needed here, such as breaking out and
+  tracking transfer or subscription notification costs separately, but only if and when it is shown that such costs are
+  a significant component of the overall cost footprint.)
+
+  Units: US$/GB
+
+Measure the above metrics on both AWS and GCP, separately and in composite (weighted average).
+
+A blend of these metrics can be thought of as the key performance indicator (KPI) for the Data Store to optimize. More
+metrics can be added or changed in each category over time, provided that observability of the system is not reduced.
```
maniarathi: Similar statement as above... instead of "can be added" maybe "should be added"?

rfcs/text/0000-dss-metrics.md
```diff
@@ -0,0 +1,78 @@
...
+
+  Units: seconds
+
+- *Error metrics*: Measure unrecoverable error rate, i.e. the rate of error that caused the client to stop trying and
+  cause a client-side error to be raised. (This may require instrumentation in the HCA CLI.)
+
+  Units: rate (number of API requests raising ultimate error (rolled up with their retries) / total API requests)
+
+- *Costing metrics*: Measure "all-in" cost per GB stored (the total cost of DSS services provided divided by the number
+  of GB stored in each replica). (Other metrics or breakout categories could be needed here, such as breaking out and
```
maniarathi: I don't think there's a need for the parenthesis for the second sentence.

maniarathi (approved):
Just nits so LGTM-ing this RFC :)

rfcs/text/0000-dss-metrics.md
kislyuk: Thanks, fixed.

rfcs/text/0000-dss-metrics.md
kislyuk: Thanks - incorporated all of these edit suggestions.

PR #90, opened 2019-07-25
Creating RFC to query by metadata schema versions so that components are notified only of data that they are compatible with.

maniarathi:
~**August 12**: Last day for community review~
~**August 23**: Last day for oversight review~
RFC approved during the 2019-08-23 Technical Architecture Team meeting.

During community review, reviewers
* corrected some errors in sample ElasticSearch queries and improved the style of others
* debated the style of whether to nest the major and minor versions of schema versions or not
* fixed some links
* discussed how to best notify teams of changes to the metadata schema
* discussed the benefits of having schema branch and revision numbers in the metadata, this was deferred to different RFC in name of moving forward with an MVP

### DIFF
--- a/rfcs/text/0000-query-by-metadata-schema-versions.md
+++ b/rfcs/text/0000-query-by-metadata-schema-versions.md
@@ -1,0 +1,239 @@
+### DCP PR:
+
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC
+number and this pull request in the format:*
+
+`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+
+# Querying DSS by Metadata Schema Version(s)
+
+
+## Summary
+
+This RFC details a design to allow downstream components to modify their subscription to the DSS to be only notified
+about bundles containing metadata within a specified schema version range. The selection of a solution is required to
+unblock the metadata schema integration testing detailed here.
+
+## Author(s)
+
+[Arathi Mani](mailto:arathi.mani@chanzuckerberg.com)
+
+## Shepherd
+
+[Matt Weiden](mailto:mweiden@chanzuckerberg.com)
+
+## Motivation
+
+Today in the DCP when metadata schema makes change, we require that the entire system, including all independent
+downstream components (i.e. Expression Matrix service, Secondary Analysis, Query service, Data Browser, etc.) are
+compatible with the changes before we allow the metadata team to push their changes through. This causes extreme
+slowness in metadata schema updates which is non-ideal due to the fact that schema changes are frequent and prevents new
+data from being added to the DCP as the new data is likely to adhere to the changed metadata schema.
+
+In order to enable more frequent metadata schema change-related pushes to production, a [metadata schema integration 
+test design](https://docs.google.com/document/d/18C_Wbrr4Huc8yxQXSdNbzeGk0glBjkfq0MGnozwlIB8/edit) has been proposed. 
+One part of design is to require downstream components to modify their production subscriptions to the Data Store 
+Service (DSS) such that they only receive notifications about bundles with which they are compatible. This will allow
+wranglers to submit new data to the DSS without having to wait for compatibility of the whole system and also allows
+downstream components to receive data independent of whether another component is ready as well. Due to the fact that
+there is currently a metadata schema freeze disallowing any changes, major, minor, or patch, to be pushed to production,
+this document simply designs the minimum system required to unblock wranglers from submitting data and protecting
+downstream components (and therefore the entire DCP) from breaking due to incompatibility.
+
+Note: the subscription modifications will only exist in the production environment. Once the metadata schema integration
+tests are setup, subscriptions will be modified to remove any restrictions on metadata schema versions in order to
+eagerly surface incompatibilities.
+
+
+### User Stories
+
+*Share the [User Stories](https://www.mountaingoatsoftware.com/agile/user-stories) motivating this RFC.*
+* As a data wrangle, I would like to be able to push new data adhering to the latest metadata schema without having to
+wait for every single component to be compatible with the new schema versions.
+* As a DCP developer working on a downstream component (e.g. Matrix Service, Secondary Analysis, etc.) I would like to
+receive notification about new data adhering to new metadata schema versions as soon as I am ready without having to
+wait for other DCP components to also be compatible. 
+
+## Detailed Design
+
+There are two questions fueling the design of this system:
+
+1) How will the schema version be noted in the metadata schema?
+2) Will ElasticSearch, JMESPath, and Query Service all be able to issue range queries against the format of the schema
+version?
+
+Today, the version of the metadata schema is embedded in the `described_by` field which contains the schema version
+embedded in the URL; an example value is “https://schema.humancellatlas.org/type/biomaterial/15.3.0/donor_organism” and
+the schema version is 15.3.0. Unfortunately embedding the version within the URL does not facilitate searching for a
+range of data between schema versions. 
+
+### New schema version fields in metadata schema
+We propose adding two new required fields to the provenance schema which is embedded into each Type schema:
+
+```
+"schema_major_version": {
+      "description": "The major version number of the schema.",
+      "type": "integer",
+      "minimum": 1,
+      "example": "4"
+},
+"schema_minor_version": {
+      "description": "The minor version number of the schema.",
+      "type": "integer",
+      "example": "6"
+}
+```
+
+We have noticeably left off adding the patch version; this is due to the fact that patch version increments only affect
+documentation about the metadata and therefore can have no effect on any code that uses the metadata schema. Since it
+does not affect forward or backwards compatibility we have chosen not to represent it explicitly in the schema which
+means that subscribing to data based on the patch version is not supported.
+
+We have decided not to represent the full version as `x.y.z` in the schema_version because formatting in this way is not
+compatible with the way that ElasticSearch performs range queries. Range queries in ElasticSearch can only be done on
+fields that are either an integer or a floating point number (or a datetime).
+
+### ElasticSearch Query for Metadata Schema Version
+```
+# Query for files where the metadata schema version of the project_json metadata schema # is greater than or equal to
+# 1.0 but less than 3.2
+query = {
+  "query": {
+    "bool": {
+      "must": {
+        "term": {
+          "files.project_json.provenance.document_id": project_uuid
+        }
+      },
+      "should": [
+        {
+          "bool": {
+            "must": {
+              "range": {
+                "files.project_json.provenance.schema_major_version": {
+                  "gte": 1,
+                  "lte": 2
+                }
+              }
+            }
+          }
+        },
+        {
+          "bool": {
+            "must": [
+              {
+                "term": {
+                  "files.project_json.provenance.schema_major_version": 3
+                }
+              },
+              {
+                "range": {
+                  "files.project_json.provenance.schema_minor_version": {
+                    "gte": 0,
+                    "lte": 2
+                  }
+                }
+              },
+            ]
+          }
+        }
+      ],
+      "minimum_should_match": 1
+    }
+  }
+}
+```
+
+Noticeably, ES queries can get long-winded very quickly especially given that there are up to 25 schemas for which
+versions can be specified. However, ElasticSearch queries are actively being deprecated, tracked by ticket 
+[here](https://app.zenhub.com/workspaces/dcp-5ac7bcf9465cb172b77760d9/issues/humancellatlas/data-store/906), and will 
+require components that currently use ElasticSearch queries for their subscriptions to switch to either JMESPath or
+Query Service.  
+
+### JMESPath Query for Metadata Schema Version
+
+```
+# Query for files where the metadata schema version of the library_preparation_protocol is greater than or equal to 1.0
+# but less than 3.2
+jmespath_query = "(event_type === `CREATE`) && ((file.library_preparation_protocol[].schema_major_version >= 1 && \
+    file.library_preparation_protocol[].schema_major_version <= 2) ||
+    (file.library_preparation_protocol[].schema_majorversion == 3 && \
+    file.library_preparation_protocol[].schema_minor_version <= 2))"
+
+```
+
+### Query-Service Query for Metadata Schema Version
+Since Query Service is currently not used as part of any downstream components’ subscriptions to the DSS, [a ticket has
+been filed](https://github.com/HumanCellAtlas/query-service/issues/221) to add functionality such that querying for data
+based on a range of metadata schema versions will be enabled in the future *by the time ElasticSearch is deprecated*.
+
+### Implementation
+The following are the implementation requirements in order to fulfill the design above. The work is split into two
+phases to enable first quickly unblocking wranglers from submitting data due to the current metadata schema freeze and
+then second the other portions of work that will be required to ensure that this design is airtight for the long term.
+
+**Phase 1 (in implementation order)**
+
+1) Add the above two required fields to metadata schemas. DO NOT make them a required field yet as all data currently in
+the DCP do not have these fields populated and therefore will be inaccessible by the downstream components if they were
+required.
+2) Make changes in Ingest such that when parsing the `describedBy` field, `schema_major_verion` and 
+`schema_minor_version` are populated as well. Add validation to make sure the values in the new fields and `describedBy`
+do not differ.
+3) Make changes in Secondary Analysis code such that when the secondary bundles are being created and the `describedBy`
+field is populated, the `schema_major_verion` and `schema_minor_version` fields are also populated with the same schema
+version information. The same validation as step 2 should trigger to make sure there is no diff in the values.
+4) **Optionally** update the subscriptions of Azul (/Data Browser), Matrix Service, Query Service, and Secondary 
+Analysis to reflect the ranges of schemas that they are compatible with. If they are agnostic to the version of a
+particular schema (for example, if they do not use the metadata schema properties in any way so its formatting doesn’t
+matter), then there is no need to specify the versions. Components may choose to opt out of specifying any versions if
+they have built-in logic to handle any errors related to metadata schema changes.
+
+Note about updating subscriptions: Components that use [metadata-api](https://github.com/humancellAtlas/metadata-api)
+must restrict the version range on every schema listed 
+[here](https://github.com/HumanCellAtlas/metadata-api/blob/master/src/humancellatlas/data/metadata/api.py#L792).
+
+At this point, new data may be submitted to the DSS by wranglers adhering to a new schema version.
+
+**Phase 2**
+
+1) Back-populate the schema versions in all existing data via either a simple script or migration tool. Re-ingest if
+needed (but hoping that this will not be necessary if [simple updates are 
+implemented](https://app.zenhub.com/workspaces/dcp-5ac7bcf9465cb172b77760d9/issues/humancellatlas/dcp/222)). 
+2) Set the two fields as now being required.
+
+At this point, the MVP of this design is complete.
+
+**Post MVP**
+
+At some point ElasticSearch will be deprecated and replaced by Query Service. Before this happens, Query Service should
+be able to handle querying by ranges of schema versions.
+
+***Future Work***
+Transitioning towards a versioning system for metadata schemas where one version represents the state of the entire set
+of metadata schemas might be a viable solution to look into by the end of the year (2019). This would prevent the
+necessity of subscriptions specifying potentially versions for multiple schemas which make the queries quite cumbersome.
+
+Another "nice-to-have" would be an automated generation of subscriptions to also lessen the complexity in drafting the
+appropriate subscriptions to adhere to potentially many schema versions. 
+
+
+### Acceptance Criteria
+
+* The current metadata schema version freeze is lifted and the metadata schema changes listed
+[here](https://docs.google.com/spreadsheets/d/1B52QjioJ7CzOLXFIcB-Q8mfjiFSH0jzbfdNs_nce3P4/edit#gid=0) are pushed to
+production.
+* Data wranglers are able to contribute data with metadata that adheres to the new schema versions.
+
+### Prior Art
+
+* The initial design was written in [a Google 
+Doc](https://docs.google.com/document/d/1BInKossmrlpna5MV_s4MSR0Bll1wnD-3OVw-9nCLdWg/edit?ts=5d389d03#) which is mostly
+a replica of the content in this RFC.
+* This work is part of a larger piece of work around introducing [metadata schema integration
+testing](https://docs.google.com/document/d/18C_Wbrr4Huc8yxQXSdNbzeGk0glBjkfq0MGnozwlIB8/edit#). The end goal beyond the
+scope of this RFC is to preemptively be able to detect breakages in the system due to metadata schema changes and
+automatically push updates when the overall DCP system is compatible. This also involves work around establishing a
+migration SOP and building out the [Metadata API](https://github.com/humancellAtlas/metadata-api).
+* [Slide deck](https://tinyurl.com/y2fljppe) presented to DCP PMs on Engineering Plan to unfreeze metadata schema 
+versions.
--- a/rfcs/text/0011-query-by-metadata-schema-versions.md
+++ b/rfcs/text/0011-query-by-metadata-schema-versions.md
@@ -1,9 +1,6 @@
 ### DCP PR:
 
-***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC
-number and this pull request in the format:*
-
-`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+[dcp-community/rfc11](https://github.com/HumanCellAtlas/dcp-community/pull/90)
 
 # Querying DSS by Metadata Schema Version(s)
 


mweiden:
@maniarathi I'm offering to shepherd here, if needed

rfcs/text/0000-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,202 @@
+### DCP PR:
+
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC
+number and this pull request in the format:*
+
+`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+
+# RFC Name
+
+Querying DSS by Metadata Schema Version(s)
```
mweiden:
```suggestion
```

rfcs/text/0000-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,202 @@
+### DCP PR:
+
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC
+number and this pull request in the format:*
+
+`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+
+# RFC Name
```
mweiden:
```suggestion
# Querying DSS by Metadata Schema Version(s)
```

rfcs/text/0000-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,202 @@
...
+
+We have decided not to represent the full version as `x.y.z` in the schema_version because formatting in this way is not
+compatible with the way that ElasticSearch performs range queries. Range queries in ElasticSearch can only be done on
+fields that are either an integer or a floating point number (or a datetime).
+
+### ElasticSearch Query for Metadata Schema Version
+```
+# Query for files where the metadata schema version of the project_json metadata schema # is greater than or equal to
+# 1.0 but less than 3.2
+query = {
```
hannes-ucsc: Not a valid ES query. Maybe it doesn't matter but I feel better if these queries were tested at least once. ES can have surprising limitations.

rfcs/text/0000-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,202 @@
...
+Today, the version of the metadata schema is embedded in the `described_by` field which contains the schema version
+embedded in the URL; an example value is “https://schema.humancellatlas.org/type/biomaterial/15.3.0/donor_organism” and
+the schema version is 15.3.0. Unfortunately embedding the version within the URL does not facilitate searching for a
+range of data between schema versions. 
+
+### New schema version fields in metadata schema
+We propose adding two new required fields to each list of schema properties:
+
+```
+"schema_major_version": {
```
hannes-ucsc: What schema file exactly in would this addition happen in? The same schema that specifies "describedBy"?

Also, consider a nested approach:

```
"schema_version": {
    "major": 1,
    "minor": 2
}

File: rfcs/text/0000-query-by-metadata-schema-versions.md [hunk 5]
```diff
@@ -0,0 +1,202 @@
...
+      "description": "The major version number of the schema.",
+      "type": "integer",
+      "pattern": "^[1-9]\\d*$",
```
hannes-ucsc: I don't think you need a pattern for integers.

rfcs/text/0000-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,202 @@
...
+# but less than 3.2
+jmespath_query = "(event_type === `CREATE`) && (file.library_preparation_protocol[].schema_major_version >= 1 && \
+    file.library_preparation_protocol[].schema_major_version <= 3 && \
+    file.library_preparation_protocol[].schema_minor_version <= 2)"
+
+```
+
+### Query-Service Query for Metadata Schema Version
+Since Query Service is currently not used as part of any downstream components’ subscriptions to the DSS, [a ticket has
+been filed](https://app.zenhub.com/workspaces/dcp-5ac7bcf9465cb172b77760d9/issues/humancellatlas/query-service/221) to
```
hannes-ucsc: I'd change this to a plain old github link, one that doesn't point at zenhub.

hannes-ucsc (changes requested):
@maniarathi and I agreed that pinning the version would be optional and I think this should be explicitly stated here. Apologies if I missed it. I also think this got moved too quickly from the Google doc to the RFC stage. There is ongoing discussion in the doc.

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: This appears in the provenance schema. https://github.com/HumanCellAtlas/metadata-schema/blob/master/json_schema/system/provenance.json

What's the benefit of the nested approach versus non-nested?

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: If this isn't filled in, the metadata linter fails.

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: Yep. I'm working on it and will update soon! :)

rfcs/text/0000-query-by-metadata-schema-versions.md
diekhans: I would agree that this is a more straight-forward representation.

diekhans (commented):
Please describe that these fields are added to system/provenance.json and the rationale being that it can be added automatically by ingest.

That plus hannes' changes and I will give this an enthusiastic thumbs up

diekhans:
It would also be good to provide a common library to implement create the subscriptions, rather than have each component recode the same thing.

rfcs/text/0000-query-by-metadata-schema-versions.md
justincc: The plan is that this will be in the provenance section of the doc.

justincc (approved):

rfcs/text/0000-query-by-metadata-schema-versions.md
malloryfreeberg: I think @maniarathi was looking to exclude having "0" be the first digit in an integer for the major version.

rfcs/text/0000-query-by-metadata-schema-versions.md
malloryfreeberg: It's also worth noting that this example only restricts 1 schema version - for `project.json` - but there are actually 25 type schemas that could potentially have restrictions on the versions for any given query.

rfcs/text/0000-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,202 @@
...
+means that subscribing to data based on the patch version is not supported.
+
+We have decided not to represent the full version as `x.y.z` in the schema_version because formatting in this way is not
+compatible with the way that ElasticSearch performs range queries. Range queries in ElasticSearch can only be done on
+fields that are either an integer or a floating point number (or a datetime).
+
+### ElasticSearch Query for Metadata Schema Version
+```
+# Query for files where the metadata schema version of the project_json metadata schema # is greater than or equal to
+# 1.0 but less than 3.2
```
calvinnhieu: The example queries for ES and JMESPath both select only a subset of versions between 1.0 and 3.2, i.e. versions with minor schema versions > 2 are excluded (e.g. 1.3, 2.4...). I believe one implementation that could work is specifying a minor version condition for every major version range included. For instance, this example would look like:

```
{
  "query": {
    "bool": {
      "must": {
        "match": {
          "files.project_json.provenance.document_id": project_uuid
        }
      },
      "should": [
        {
          "bool": {
            "must": {
              "range": {
                "files.project_json.provenance.schema_major_version": {
                  "gte": 1
                  "lte": 2
                }
              }
            }
          }
        },
        {
          "bool": {
            "must": [
              {
                "match": {
                  "files.project_json.provenance.schema_major_version": 3
                }
              },
              {
                "range": {
                  "files.project_json.provenance.schema_minor_version": {
                    "gte": 0,
                    "lte": 2
                  }
                }
              },
            ]
          }
        }
      ],
      "minimum_should_match": 1
    }
  }
}
```

If there's a simpler way to express AND and OR clauses in ES, I'd be glad to hear it!

calvinnhieu: Also, I'm assuming downstream components defining their subscription ES queries will potentially have to specify a similar `should` clause to the above for every metadata file we want to control. While I doubt this will affect performance, I will note that queries will become relatively long/verbose. Granted, this sounds like a short term solution before integration tests are implemented, so I'm OK with this.

rfcs/text/0000-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,202 @@
...
+eagerly surface incompatibilities.
+
+
+### User Stories
+
+*Share the [User Stories](https://www.mountaingoatsoftware.com/agile/user-stories) motivating this RFC.*
+* As a data wrangle, I would like to be able to push new data adhering to the latest metadata schema without having to
+wait for every single component to be compatible with the new schema versions.
+* As a DCP developer working on a downstream component (e.g. Matrix Service, Secondary Analysis, etc.) I would like to
+receive notification about new data adhering to new metadata schema versions as soon as I am ready without having to
```
calvinnhieu: Maybe more of a question for the metadata team, but as this RFC outlines a solution for downstream services to control updates to specific metadata versions, what is the best venue for developers to stay up-to-date with the latest metadata schema versions and, to their discretion, respond and update their services to support newer versions?

calvinnhieu (changes requested):
Update example queries to reflect intended use case, but otherwise, sounds great to me!

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: Correct. Resolving this conversation as there is no action item here.

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: I will have to run this by the metadata folks... @simonjupp what do you think?

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: Woah! Thank you for the query @calvinnhieu !!

Yes, there is ongoing work to deprecate ES queries so this should be a short-term solution.

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: Done.

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: I added a note below the ES query about its long-winded-ness and also the fact that there are 25 schemas. I also linked the ticket towards the deprecation of ES so it should be be much simpler to write subscriptions moving forward with either JMESPath or Query Service. Please feel free to re-open if there are lingering issues.

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: Fixed both the ES and JMESPath queries. Thank you so much!

maniarathi:
> It would also be good to provide a common library to implement create the subscriptions, rather than have each component recode the same thing.

@diekhans This is definitely something I consider a "nice-to-have" rather than a feature to stick in the MVP :) I added a section to the end of the RFC on Future Work and included the automatic generation there.

rfcs/text/0000-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,237 @@
...
+      "type": "integer",
+      "pattern": "^[1-9]\\d*$",
+      "example": "4"
+},
+"schema_minor_version": {
+      "description": "The minor version number of the schema.",
+      "type": "integer",
+      "pattern": "^\\d+$",
+      "example": "6"
+}
```
xbrianh: I'd like to propose a third field in addition to `schema_major_version` and `schema_minor_version`: `schema_revision_number`, which could be any increasing function of schema changes. For instance, the number of commits to [metadata-schema](https://github.com/HumanCellAtlas/metadata-schema) would probably work, which can be found with the command `git rev-list --count HEAD`.

This would provide a way to pin a large swath of metadata.

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: Great question! This is definitely a discussion worth having.. 

From my perspective, once the metadata schema integration tests are implemented, there will be notifications sent out on "how your component broke with the latest schema changes." 

But a general digest might also make sense. @simonjupp , @malloryfreeberg , @morrisonnorman ... any thoughts?

rfcs/text/0000-query-by-metadata-schema-versions.md
samanehsan: @xbrianh would the `schema_revision_number` only increment for the files that get changed (so the number only has significance within the context of an individual file)? Or would it function more like a snapshot of all the files (kind of like a commit hash)?

xbrianh: @samanehsan More like a snapshot of all of the files.

samanehsan: Thanks for the clarification! That would be incredibly valuable to us since analysis could just pin to a particular snapshot.

xbrianh (approved):

calvinnhieu (approved):
Looks good!

maniarathi:
@diekhans Modified to add the note about the schema versions being placed into the provenance block.

kislyuk (approved):

rfcs/text/0000-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,237 @@
...
+***Future Work***
+Transitioning towards a versioning system for metadata schemas where one version represents the state of the entire set
+of metadata schemas might be a viable solution to look into by the end of the year (2019). This would prevent the
+necessity of subscriptions specifying potentially versions for multiple schemas which make the queries quite cumbersome.
+
+Another "nice-to-have" would be an automated generation of subscriptions to also lessen the complexity in drafting the
+appropriate subscriptions to adhere to potentially many schema versions. 
+
+
+### Acceptance Criteria [optional]
```
mweiden:
```suggestion
### Acceptance Criteria
```

rfcs/text/0000-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,237 @@
...
+
+*Acceptance criteria are the conditions that a RFC must satisfy to be accepted by users or other stakeholders.*
```
mweiden:
```suggestion
```

rfcs/text/0000-query-by-metadata-schema-versions.md
hannes-ucsc: I strongly urge the author to consider the addition of a monotonically increasing revision number as @xbrianh suggests. If we postpone this decision, we'd have to rewrite the metadata again once it is adopted.

The `git rev-list --count HEAD` command could produce the same number for two distinct branches with the same number of commits but with different commits in the divergent parts of the those branches. Adding a forth property, `schema_branch`, could be used to disambiguate them.

rfcs/text/0000-query-by-metadata-schema-versions.md
hannes-ucsc: @maniarathi in all fairness, you should leave resolution to the original author of the comment. They might not agree with your assessment that there is no action here.

hannes-ucsc: Yes, I missed the intent. Thanks for pointing that out, @malloryfreeberg.

Wouldn't 

```suggestion
      "minimum": 1,
```

be a more straightforward way to achieve this?

rfcs/text/0000-query-by-metadata-schema-versions.md
malloryfreeberg: Who would populate the `schema_revision_number` and `schema_branch` fields? Ingest for primary submissions; data processing pipelines for secondary submissions?

rfcs/text/0000-query-by-metadata-schema-versions.md
malloryfreeberg: We have tried various tactics for communicating schema changes: Posting "notes" on Slack (which seemed people rarely read and were a 100% manual thing to maintain); Integration of metadata schema PRs with the #hca-metadata-int Slack channel (what we do now, don't know if people find it useful); Tagging individual people in tickets/PRs for changes in fields that we know they hard-code against (not sustainable; requires too much knowledge). 

Would love for the DCP to find a way to improve transparency of schema changes without adding manual work for schema developers. Do any other components communicate their code changes to the rest of the DCP? If so, does their communication system work? Maybe we can adapt things that work well for other teams :)

rfcs/text/0000-query-by-metadata-schema-versions.md
malloryfreeberg: I think it's fine to have schema version fields in provenance. We want to ultimately move `describedBy` there, as well.

malloryfreeberg (approved):
I have some outstanding questions about the `schema_revision_number` and `schema_branch` fields, but I trust everyone else to reach a decision after I leave. So, happy to approve!

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: @hannes-ucsc  1) Apologies for resolving it too quickly. It was simply a habit from my previous job where folks re-opened if necessary. Simply a difference in process :) 2) minimum sounds reasonable to me. I will commit the change.

justincc: Frankly I agree 100% with closing and re-opening but this seems to upset people. Instead  stuff just gets ignored because few people come back and close their comments. Personally I prefer a transparent close but :man_shrugging:

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: Great response Mallory! I think today the DCP-wide tests that occur in integration, often surface errors between components and the metadata schema integration tests will help with that. We could try that out first and see if we need anything additional once that work is completed!

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: I think the question was about the nested versus non-nested approach to specifying the schema version. Definitely will be in the provenance block and I made a note as well.

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: @hannes-ucsc I totally understand your perspective and I agree with you that there are a lot of benefits to having an overall version but I cannot agree to adding this at this moment. There are pros and cons to an overall schema which are much more nuanced that this comment thread, not to mention that this is not a decision I can make in a vacuum, and I'm doing my absolute best to consider this option by a) having received permission from PMs/POs to schedule a Spike *in Q3* (https://app.zenhub.com/workspaces/dcp-5ac7bcf9465cb172b77760d9/issues/humancellatlas/dcp/465), b) scheduling the meeting next week with the key stakeholders, and c) committing to writing a pre-RFC document on the pros/cons of having an overall schema version AND potentially ideas on how to compute this/store this and its implications.

I have spent so many hours trying to get to a first MVP that *covers a lot, but not all* use cases and this is finally one that almost everyone has agreed to run with. To your point about having to rewrite metadata again, this is somewhat moot to me once we get migrations rolling since that is what will happen on a regular basis anyways for data releases.

I am asking you to, please, meet me halfway and agree to at least storing the major/minor schema version for now knowing that I have, and I will continue, put in a ton of time trying to expedite the conversation about an overall schema.

hannes-ucsc: Could Ingest do it for both?

But thinking more about this, especially adding the branch reference seems difficult to do. Still, the idea of a monotonically increasing revision number, even if not derived from the # of commits in the git branch would be useful. The revision number could be put into schemas when they are published on `schema.humancellatlas.org` and copied into each document by ingest when a document is created.

Or maybe the revision number would just be present in one schema/document pair only: `project.json` has always had a special role. It is currently not linked into the graph and its mere presence in a bundle associates all other documents in that bundle with the project. The revision number could be placed only into the project schema and only copied into `project.json` when that file is created. Client would then have the choice to pin the semantic version of individual documents or the monotonically increasing revision number on `project.json` only. Or a range of either.

maniarathi: @hannes-ucsc Pardon, but what do you mean by "for both"?

I don't disagree with you at all that supporting a monotonically increasing revision number is useful. I agree with you! I also think you bring some really great ideas for implementing this to the table and I am definitely going to add all of the comments above to the document I am preparing for next week's meeting.

The only thing I am asking is to at least let the ball start rolling with this proposed MVP. I am 100% committing to iterating on this design quickly.

claymfischer: If we allow scope creep we may never be ready. _Release then iterate_ makes sense to me.

justincc: Yeah, frankly there are already situations where we need to go back and change older bundles (fixing bugs, inserting entities) and I don't see that changing. We can do this iteratively.

A cheap way of doing it on your end might be to concat major and minor version number, as long as it isn't possible for there to be branches where an older schema version still gets a minor version update after there's some separate new major version. But if this was the case a monotonic revision wouldn't be of use anyway.

samanehsan (approved):

rfcs/text/0000-query-by-metadata-schema-versions.md
hannes-ucsc: > @hannes-ucsc Pardon, but what do you mean by "for both"?

For both primary and secondary bundles.

> The only thing I am asking is to at least let the ball start rolling with this proposed MVP. I am 100% committing to iterating on this design quickly.

Sounds good. I'll rest my case on this issue. 

Punting this does mean that Analysis has to pin essentially every schema referenced here:

https://github.com/HumanCellAtlas/metadata-api/blob/11480262abe484facebee016ce7e63a366c8743a/src/humancellatlas/data/metadata/api.py#L792

@justincc 

> A cheap way of doing it on your end might be to concat major and minor version number

Wouldn't 1.10 and 11.0 would produce the same value? `int(major) * 100 + int(minor)` might be an alternative.

Anyways, not only does the revision number need to be monotonically increasing, we need one revision number for the entire schema!

rfcs/text/0000-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,236 @@
...
+                }
+              }
+            }
+          }
+        },
+        {
+          "bool": {
+            "must": [
+              {
+                "match": {
```
hannes-ucsc:
```suggestion
                "term": {
```

rfcs/text/0000-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,236 @@
...
+"schema_major_version": {
+      "description": "The major version number of the schema.",
+      "type": "integer",
+      "minimum": 1,
+      "example": "4"
+},
+"schema_minor_version": {
+      "description": "The minor version number of the schema.",
+      "type": "integer",
+      "pattern": "^\\d+$",
```
hannes-ucsc:
```suggestion
```

rfcs/text/0000-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,236 @@
...
+Service (DSS) such that they only receive notifications about bundles with which they are compatible. This will allow
+wranglers to submit new data to the DSS without having to wait for compatibility of the whole system and also allows
+downstream components to receive data independent of whether another component is ready as well. Due to the fact that
+there is currently a metadata schema freeze disallowing any changes, major, minor, or patch, to be pushed to production,
+this document simply designs the minimum system required to unblock wranglers from submitting data and protecting
+downstream components (and therefore the entire DCP) from breaking due to incompatibility.
+
+Note: the subscription modifications will only exist in the production environment. Once the metadata schema integration
+tests are setup, subscriptions will be modified to remove any restrictions on metadata schema versions in order to
+eagerly surface incompatibilities.
```
hannes-ucsc:
```suggestion
eagerly surface incompatibilities.

Note: A component may opt out of implementing this change, if that component fails atomically when processing a notification for a bundle referencing a schema the component is incompatible with.

Note: Components that use the metadata-api must restrict the version range on every schema listed [here](https://github.com/HumanCellAtlas/metadata-api/blob/11480262abe484facebee016ce7e63a366c8743a/src/humancellatlas/data/metadata/api.py#L792).

```

rfcs/text/0000-query-by-metadata-schema-versions.md
hannes-ucsc: This is a nit for me and my approval does not depend on it. I think the nesting makes sense and should be easy to implement. It would prevent the proliferation of related `schema_…` properties in the provenance section, for example if we add the revision number in the future. If this nit is addressed, the example queries below would have to be adjusted (`s/schema_/schema.`).

hannes-ucsc (changes requested):

rfcs/text/0000-query-by-metadata-schema-versions.md
justincc: @hannes-ucsc Apologies  - I'm an idiot and I wrote that late at night. I realized in the car that my proposal wouldn't work.

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: Resolving conversation for now.. please feel free to reopen if everything is not addressed!

rfcs/text/0000-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,235 @@
...
+
+### ElasticSearch Query for Metadata Schema Version
+```
+# Query for files where the metadata schema version of the project_json metadata schema # is greater than or equal to
+# 1.0 but less than 3.2
+query = {
+  "query": {
+    "bool": {
+      "must": {
+        "match": {
```
hannes-ucsc:
```suggestion
        "term": {
```

rfcs/text/0000-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,235 @@
...
+          "files.project_json.provenance.document_id": project_uuid
+        }
+      },
+      "should": [
+        {
+          "bool": {
+            "must": {
+              "range": {
+                "files.project_json.provenance.schema_major_version": {
+                  "gte": 1
```
hannes-ucsc:
```suggestion
                  "gte": 1,
```

rfcs/text/0000-query-by-metadata-schema-versions.md
hannes-ucsc:
```suggestion
        "term": {
```

hannes-ucsc (changes requested):

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: I already had the first note written later on in the RFC in the implementation section so I added the second note onto it.

rfcs/text/0000-query-by-metadata-schema-versions.md
maniarathi: Ok sounds good! Still working on getting verification. The specific part I am unsure about is whether having nested properties means that I would have to create an entire new schema for the schema_version property. From pretty much every other example I've looked at, it seems as though I would have to create an entirely new schema_version.json schema which I would rather not do.

hannes-ucsc: Technically you can nest the schemas for nested objects but I don't know if that's against the MD teams's rules. Even if it were, adding a module (that's what they call schemas for nested objects) would allow it to be reused in other places. And it would not be hard to do. I'll file the MD PR if that helps.

hannes-ucsc (approved):

morrisonnorman (approved):

mweiden:
RFC approved during the 2019-08-23 Technical Architecture Team meeting.

rfcs/text/0011-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,236 @@
...
+
+Note: the subscription modifications will only exist in the production environment. Once the metadata schema integration
+tests are setup, subscriptions will be modified to remove any restrictions on metadata schema versions in order to
+eagerly surface incompatibilities.
+
+
+### User Stories
+
+*Share the [User Stories](https://www.mountaingoatsoftware.com/agile/user-stories) motivating this RFC.*
+* As a data wrangle, I would like to be able to push new data adhering to the latest metadata schema without having to
```
sampierson, line 48: s/wrangle/wranger/

rfcs/text/0011-query-by-metadata-schema-versions.md
```diff
@@ -0,0 +1,236 @@
...
+test design](https://docs.google.com/document/d/18C_Wbrr4Huc8yxQXSdNbzeGk0glBjkfq0MGnozwlIB8/edit) has been proposed. 
+One part of design is to require downstream components to modify their production subscriptions to the Data Store 
+Service (DSS) such that they only receive notifications about bundles with which they are compatible. This will allow
+wranglers to submit new data to the DSS without having to wait for compatibility of the whole system and also allows
+downstream components to receive data independent of whether another component is ready as well. Due to the fact that
+there is currently a metadata schema freeze disallowing any changes, major, minor, or patch, to be pushed to production,
+this document simply designs the minimum system required to unblock wranglers from submitting data and protecting
+downstream components (and therefore the entire DCP) from breaking due to incompatibility.
+
+Note: the subscription modifications will only exist in the production environment. Once the metadata schema integration
```
sampierson, line 40: I don't believe this mechanism should only run in production. 
Integration should eagerly surface incompatibilities, sure.  Chaos is more acceptable there.
However, the purpose of staging is to _replicate as closely as possible the production environment_, and be our "smoke test" before production deployment.  This means it should bias towards stability, and use, as closely as possible, all the mechanisms that production uses.

hannes-ucsc: +1

PR #91, opened 2019-07-26
AUDR analysis updates RFC

justincc:
~~**August 12**: Last call for community review~~
**August 23**: Last day for oversight review

During community review, reviewers

- Discussed requirements around updating/deleting files in envelopes through ingest api as a part of the process of updating analysis bundles
- Small updates to wording and titles


rfcs/text/0000-audr-analysis-updates.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Detailed Design
+A primary metadata update triggers a datastore notification. Per pipeline, Analysis performs:
+
+* **Update analysis.** The [meta]data change is in a file or parameter that updates an existing analysis (a “consequential” change).
+* **New analysis.** The [meta]data change creates a new analysis. For example, the species changes from mouse to human and there is a pipeline for human on this data but not one for mouse. It may be that an analysis from a different pipeline already exists and this should not be superseded by the new analysis.
+* **No update.** There is an existing analysis for this pipeline but the [meta]data change does not alter the analysis (an “inconsequential” change). For example, a change in the project description.
+
+The **update analysis** and **new analysis** scenarios can occur without a primary update. For example, a pipeline bug may be fixed which requires updates of the analyses it performed. Or a new pipeline may be created that can be run on existing data, where the existing analyses from other pipelines are still valid and so should not be superseded.
+
```
diekhans, line 41: It would be good to explicitly state that analysis is responsible for determining if an update is consequential.

rfcs/text/0000-audr-analysis-updates.md
```diff
@@ -0,0 +1,118 @@
...
+1. Create a submission envelope with the parameter isUpdate = true.
+2. Add `analysis_process_0.json` with provenance.document_id set to the same UUID as the existing `analysis_process_0.json`
+3. For other files that need updating, also add them to the envelope with the same provenance.document_id as their previous versions`.`
+4. For files that only exist in the new analysis, add them to the envelope with any provenance.document_id.
+5. For files that exist only in the older analysis, post an HTTP DELETE to their endpoint.
+6. For files that are in both envelopes but don’t need updating, do nothing.
+
+![](../images/0000-update-file-relationships.png)
+
```
diekhans, line 63: it would be good to describe that updated files are not really updated but deleted and then an independent new file is created and why.  Also that this is a short term compromise to simplify implementation at the expense of a more complex data mode.

diekhans: it appears this is no longer the case and the figure just needs updated.

diekhans (changes requested):
looking really good; clear and to the point.  images are great.  just minor updates requested.

rfcs/text/0000-audr-analysis-updates.md
justincc: For an updated analysis the figure tries to show the possible scenarios of a) not changing a file, b) updating a file, c) adding a file, d) deleting a file. But it sounds like this isn't easy to see and I should improve the diagram?

On the content, yes on further consideration deleting and re-adding all the other files seems a worse idea since it means some special case work and I'm not sure it ultimately helps analysis much. This is a change from what was discussed so I need feedback from @rexwangcc and @samanehsan.

samanehsan: @justincc can you clarify why the old files need to be deleted in order to add analysis files that have new uuids?

rfcs/text/0000-audr-analysis-updates.md
```diff
@@ -0,0 +1,118 @@
+### DCP PR:
+
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC
+number and this pull request in the format:*
+`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+
+# RFC Name
+AUDR Analysis Updates
```
mweiden:
```suggestion
```

rfcs/text/0000-audr-analysis-updates.md
```diff
@@ -0,0 +1,118 @@
+### DCP PR:
+
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC
+number and this pull request in the format:*
+`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+
+# RFC Name
```
mweiden:
```suggestion
# AUDR Analysis Updates
```

rfcs/text/0000-audr-analysis-updates.md
justincc: @samanehsan Because everything in the previous version of that bundle will automatically carry forward.

rfcs/text/0000-audr-analysis-updates.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Motivation
+This covers [use cases](https://docs.google.com/document/d/1rI8PCASomdAHznyWQceRJGv4-wtg-P5Rm-rI_uJLjtE/edit#):
+
+* **Fix an incorrect metadata field (e.g. cell count)** where the change can trigger re-analysis.
+* **Fix an incorrect ontology term** where the change can trigger re-analysis.
+* Fix incorrect analysis using a corrected pipeline.
+* **Add better analysis using an improved pipeline.**
+
+Wranglers consider the **bolded** use cases in red to be critical for general availability.
```
parthshahva: Can we remove the reference to GA? We are already technically in "GA" and this label means different things to different people.

parthshahva: Can we remove the reference to GA? We are already technically in "GA" and this label means different things to different people.

rfcs/text/0000-audr-analysis-updates.md
parthshahva: @justincc I'm a little confused as to why the old files from the original analysis bundle need to be deleted. Why would you copy forward files from the old analysis bundle to the new analysis bundle? Shouldn't the copy forward be from the new primary bundle to the new analysis bundle?

rfcs/text/0000-audr-analysis-updates.md
justincc: Done

rfcs/text/0000-audr-analysis-updates.md
justincc: Sorry, my use of 'carry forward' was confusing - I didn't mean the same thing as 'copy forward'.

When an update comes in, ingest assumes that all the files in the previous version of that bundle should also be in the new version unless told otherwise. This is the way it has to work for primary submissions since an update can't realistically respecify all files that haven't changed. So it ends up working the same way for secondary updates - we will assume all files in the previous analysis bundle should also be in the new unless told otherwise.

samanehsan: Thanks for the clarification @justincc. Will the copied analysis files have the same uuid and a new version? In that case, could we "update" all of the analysis files even if some of them haven't changed?

justincc: In this case they would have the same version. I've added this to the text.

parthshahva (approved):

kislyuk (approved):

diekhans (approved):
great work!

rfcs/text/0000-audr-analysis-updates.md
parthshahva: Thanks!

rfcs/text/0000-audr-analysis-updates.md
samanehsan: I was discussing this further with @rexwangcc and we found a potential complication. Because each bundle is independently processed via a Cromwell workflow, it does not have any knowledge of what else is being processed. When a workflow submits an analysis update, we would make a request to the data store to retrieve the latest analysis bundle and compare files between bundles. However, if another workflow submitted an analysis update to the same bundle but it is not yet in the data store, analysis would end up using the wrong analysis bundle version to compare file changes. This can happen because there are several steps between analysis creating a submission in ingest, and an analysis bundle appearing in the data store.

@justincc is there a way that ingest could determine which files are new/updated/deleted right before creating the new analysis bundle?

parthshahva: @samanehsan @rexwangcc  When would you have multiple workflows of the same pipeline running parallel updates on the same analysis bundle? If you had multiple "types" of pipelines running, wouldn't each pipeline lineage have a separate bundle UUID? The flow you discuss sounds like a notification/ workflow triggering bug rather than a scenario to account for

samanehsan: @parthshahva, this scenario is if two consequential updates occur close together. So the pipelines are still triggered sequentially, but if the first update workflow is still staging files or waiting for the submission to go valid when the second update workflow is trying to update the "latest" submission in ingest, the second workflow doesn't know about the one that is already in progress - it would just check what is complete in the data store.

justincc: @samanehsan Okay, so if Analysis can't wait until the first submission is complete this is where Ingest will have to do a major chunk of engineering instead. Ingest doesn't currently have any protection against simultaneous submissions interleaving updates. This will likely be a very hard problem to address. The simplest solution may be to only ever process one submission at a time but that would be a permanent brake on throughput.

At some point Ingest needs to handle this problem anyway since a primary submitter may also make 2 submissions for the same data in quick succession. But at the moment the wranglers can control that so we were hoping to kick this can down the road.

Could you confirm that 2 analyses triggering in quick succession is a scenario that's not sufficiently theoretical for us to ignore temporarily?

justincc: Regarding the actual submission, you submit update documents. It is Ingest that calculates the delta from its own records. If Ingest processes the submissions strictly sequentially you would be fine.

samanehsan: > Could you confirm that 2 analyses triggering in quick succession is a scenario that's not sufficiently theoretical for us to ignore temporarily?

@justincc I don't think the scenario is unrealistic because it could happen if a primary submitter makes/updates a submission even an hour or two apart. Looking at this [production workflow](https://job-manager.caas-prod.broadinstitute.org/jobs/82f72ecb-8ce9-496c-b321-66c2cf857f33?q=caas-collection-name%3Dlira-prod%26status%3DSucceeded) as an example, after analysis creates a submission envelope in ingest it takes 2 hours to upload the files and confirm the submission. So if an update came in less than 2 hours later, we would have this "simultaneous" submission problem.

Also in this case, the confirm submission step only took 4 minutes but the workflow can wait up to 2 hours for the submission to validate. And then there is additional time for the submission to go through ingest and appear as a results bundle in the data store. So in the worst-case scenario, "quick succession" actually means 2+ hours.

rfcs/text/0000-audr-analysis-updates.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Motivation
+This covers [use cases](https://docs.google.com/document/d/1rI8PCASomdAHznyWQceRJGv4-wtg-P5Rm-rI_uJLjtE/edit#):
+
+* **Fix an incorrect metadata field (e.g. cell count)** where the change can trigger re-analysis.
+* **Fix an incorrect ontology term** where the change can trigger re-analysis.
+* Fix incorrect analysis using a corrected pipeline.
+* **Add better analysis using an improved pipeline.**
+
+Wranglers consider the **bolded** use cases in red to be of greater importance than the other use cases.
```
sampierson: "in red"?  I don't see red when I view the file.

justincc: corrected

rfcs/text/0000-audr-analysis-updates.md
samanehsan: @justincc, this is the concern that @rexwangcc and I mentioned in tech-arch today ^

justincc: Yes, I think the answer here is that ingest may need to do complex engineering to ensure non-interference between submissions. Does that address the issue?

samanehsan: Can you clarify what that will look like in ingest? Is the idea that analysis would still specify which files we think are new/updated/deleted and then ingest confirms? That seems like we would be duplicating effort.

justincc: At the moment there's no plan for ingest to confirm anything, it just does what you say or some known interpretation thereof. If it's an update then the new update (which is given as the file, not a delta) will replace whatever the old version was. If it's addition of a file that already exists then it could just replace what is already there if this is the easiest behaviour. Or if it's a deletion of something that's already deleted then we could ignore that if this is the easiest behaviour.

samanehsan:
@justincc there's an "rfc-paused" label you can add as well

PR #88, opened 2019-07-22
Processing Datasets that Span Multiple Data Collection Runs

barkasn:
This RFC proposes a design solution to allow datasets that span multiple data collection runs to be processed in unison as per the scientific requirements. The proposal generalizes to multiple data type modalities, but also provides specific details for 10X V2 3’ single-cell data as an example and to address an immediate need for a data processing solution.

This RFC has undergone a substantial rewrite based on comments.

Summary of Discussion for **_Approvers_**:
This RFC initially proposed the concept of DAPS as a design solution to allow datasets that span multiple data collection runs to be processed in unison. After an extensive rewrite the concept of **data groups** is now proposed. A **data group** is a collection of bundles that are version-complete, that is at a given point in time fulfill specified completeness criteria. An implementation of the **data group** concept as a bundle is proposed in order to re-use existing infrastructure. The completeness criteria that the **data group** must fulfill depend on its type. This RFC proposes the implementation of the PROJECT_SUBMISSION **data group type**, that specifies that a submission of a data set is complete and downstream processing can commence. Other **data group** types can be added in the future as project needs dictate.

~~**August 27th, 2019** Last Call for Community Review~~
**September 27th, 2019** Last call for oversight review

PR #93, opened 2019-07-29
RFC: HCA DCP Bundle Types and Definitions

malloryfreeberg:
Please note: the contents of #86 were merged into this PR; please refer to the discussion in that PR for additional detail and comments addressed from reviewers.

### DIFF
--- a/rfcs/text/0000-bundle-definition.md
+++ b/rfcs/text/0000-bundle-definition.md
@@ -73,19 +73,19 @@ In the HCA DCP, a bundle is defined as:
 Bundle types, definitions, consumers, and specifications are outlined below. Use cases from above are indicated where supported by a specific bundle type.
 
 Definitions:
-**Type**: The high-level type of bundle (See Bundle Type RFC)
-**Definition**: What is and is not in this bundle type
-**Target consumer**: Who will be consuming this bundle
-**Use cases**: Which Users Stores from above are supported by this bundle type
-**Specification**: Details about the bundle type
+- **Type**: The high-level type of bundle (See Bundle Type RFC)
+- **Definition**: What is and is not in this bundle type
+- **Target consumer**: Who will be consuming this bundle
+- **Use cases**: Which Users Stores from above are supported by this bundle type
+- **Specification**: Details about the bundle type
 
-| Type | Definition | Target consumer | Use cases | Detailed specification |
+| Type | Definition | Target consumer | Use cases | Specification |
 |:-|:-|:-|:-|:-|
 | Primary bundle | A bundle that contains primary (raw) data files and all related metadata files. | computational biologists; Data Processing Pipelines; Data Browser; Query Service | 1 | Sequencing: type = sequence; Imaging: type = image |
 | Secondary bundle | A bundle that contains secondary (alignment, expression) data files produced by the Data Processing Pipelines, input primary data files, and all related metadata files. | computational biologists; Data Browser; Query Service; Matrix Service | 2, 3 | Sequencing: type = [alignment, expression]; Imaging: type = [expression, ?] |
 | Tertiary bundle | A bundle that contains tertiary (expression) data files produced by the Matrix Service, input primary and secondary data files, and all related metadata files. | computational biologists; biologists; Data Browser; Query Service | 2, 3 | type = [expression, ?] |
 | Project bundle | A bundle that contains project-level metadata files. This bundle does not contain data files. | computational biologists; biologists; | 1, 2, 3, 4 | ? |
-| Resource/Reference bundle | A bundle that contains reference data files and all related metadata files. | computational biologists; Data Processing Pipelines; | 4 | ? |
+| Resource/Reference bundle | A bundle that contains reference data files used by the Data Processing Pipelines(?) and all related metadata files. | computational biologists; Data Processing Pipelines; | 4 | ? |
 
 
 
--- a/rfcs/text/0000-bundle-definition.md
+++ b/rfcs/text/0000-bundle-definition.md
@@ -48,7 +48,7 @@ There are currently no documented definitions for what a "bundle" is in the HCA 
 *Share the [User Stories](https://www.mountaingoatsoftware.com/agile/user-stories) motivating this RFC.*
 
 1. As a data consumer who develops data processing tools (*e.g.* a sequence read aligner), I would like to browse, query, and get primary data from the DSS so that I can use it as testing data during development.
-1. As a data consumer who develops analysis tools (*e.g.* a clustering algorithm), I would like to browse, query, and get processed data from the DSS so that I can use it as testing data during development.
+1. As a data consumer who develops analysis tools (*e.g.* a clustering algorithm), I would like to browse, query, and get alignment data from the DSS so that I can use it as testing data during development.
 1. As a data consumer who is a computational biologist, I would like to get expression matrices from the DSS so that I can compare against data I already have.
 1. As a data consumer who is a biologist, I would like to get expression matrices from the DSS so that I can see the expression profile of <my gene of interest> in <cell type(s) or organ(s) of interest>.
 1. As a data consumer who is a computational biologist, I would like to know what reference data was used to generate processed data in the DSS and I would like to be able to get it for myself so that I can compare with my own research.
--- a/rfcs/text/0000-bundle-definition.md
+++ b/rfcs/text/0000-bundle-definition.md
@@ -8,76 +8,118 @@
 
 ## Summary
 
-There are currently no documented definitions for what a "bundle" is in the HCA DCP, including no definitions of what they are, what they contain, or who they support. Lack of clarity causes confusion for data consumers about how the data and metadata in the DCP are structured and organized. Here we propose to formalize a bundle definition, bundle type definitions, target users and consumers, and specific use cases in order to support the needs of data consumers.
+There are currently no documented definitions for what a "bundle" is in the HCA DCP, including no definitions of what
+they are, what they contain, or who they support. Lack of clarity causes confusion for data consumers about how the data
+and metadata in the DCP are structured and organized. This RFC formalizes *de facto* data bundle types, definitions,
+target users/consumers, and specific use cases to improve clarity and confidence among DCP developers when working with
+the DCP data model.
 
 Specifically, the following charters will be supported by this RFC:
-- the Data Store (DSS) [charter](https://github.com/HumanCellAtlas/dcp-community/blob/master/charters/DataStore/charter.md#core-capabilities) ("The Data Store owns the bundle use cases and bundle type definitions, while the precise specifications will be negotiated between the Metadata and other teams.");
-- the Ingestion Service [charter](https://github.com/HumanCellAtlas/dcp-community/blob/master/charters/IngestionService/charter.md#dependencies) ("The Ingestion Service has a dependency on the Data Store team to provide new bundle requirements, and the Metadata Schema team to provide the definition of bundle specifications, in order to support the generation of bundles from submitted data and metadata."); and
-- the Metadata Schema [charter](https://github.com/HumanCellAtlas/dcp-community/blob/master/charters/MetadataSchema/charter.md#objectives) ("Working with the Data Store to receive new bundle structure requirements and defining a new specification to be implemented by the Ingestion Service team."). 
 
+- The Data Store (DSS)
+  [charter](https://github.com/HumanCellAtlas/dcp-community/blob/master/charters/DataStore/charter.md#core-capabilities)
+  ("The Data Store owns the bundle use cases and bundle type definitions, while the precise specifications will be
+  negotiated between the Metadata and other teams.");
+
+- The Ingestion Service
+  [charter](https://github.com/HumanCellAtlas/dcp-community/blob/master/charters/IngestionService/charter.md#dependencies)
+  ("The Ingestion Service has a dependency on the Data Store team to provide new bundle requirements, and the Metadata
+  Schema team to provide the definition of bundle specifications, in order to support the generation of bundles from
+  submitted data and metadata."); and
+
+- The Metadata Schema
+  [charter](https://github.com/HumanCellAtlas/dcp-community/blob/master/charters/MetadataSchema/charter.md#objectives)
+  ("Working with the Data Store to receive new bundle structure requirements and defining a new specification to be
+  implemented by the Ingestion Service team.").
 
 ## Author(s)
 
-*Recommended format for Authors:*
-
- [Mallory Freeberg](mailto:mfreeberg@ebi.ac.uk)
+* [Mallory Freeberg](mailto:mfreeberg@ebi.ac.uk) (TODO: Replace Mallory with Zina P and/or Chris V.)
  
- [Brian Hannafious](mailto:bhannafi@ucsc.edu)
+* [Brian Hannafious](mailto:bhannafi@ucsc.edu)
  
- [Hannes Schmidt](mailto:hannes@ucsc.edu)
+* [Hannes Schmidt](mailto:hannes@ucsc.edu)
  
- [Andrey Kislyuk](mailto:akislyuk@chanzuckerberg.com)
-
-> Replace Mallory with Zina P and/or Chris V.
+* [Andrey Kislyuk](mailto:akislyuk@chanzuckerberg.com)
 
 ## Shepherd
-***Leave this blank.** This role is assigned by DCP PM to guide the **Author(s)** through the RFC process.*
 
-*Recommended format for Shepherds:*
-
- `[Name](mailto:username@example.com)`
+* [Andrey Kislyuk](mailto:akislyuk@chanzuckerberg.com)
 
 ## Motivation
 
-*Describe the user or technical need in detail [with alignment to the DCP roadmap priorities where possible]. Link prior community discussions to demonstrate support for this RFC.*
-
-There are currently no documented definitions for what a "bundle" is in the HCA DCP, including no definitions of what they are, what they contain, or who they support. Lack of clarity causes confusion for data consumers (and DCP developers) about how the data and metadata in the DCP are structured and organized. At the June 2019 DCP F2F meeting in Cambridge, a breakout session was held to discuss "What is a bundle?", and a lot of useful discussion occurred. This RFC will formally document what a bundle is, anchoring the definition on how bundles support HCA DCP data consumers. This RFC will also describe new bundles types that could be supported by the HCA DCP.
+There are currently no documented definitions for what a "bundle" is in the HCA DCP, including no definitions of what
+they are, what they contain, or who they support. Lack of clarity causes confusion for data consumers (and DCP
+developers) about how the data and metadata in the DCP are structured and organized. At the June 2019 DCP F2F meeting in
+Cambridge, a breakout session was held to discuss "What is a bundle?", and a lot of useful discussion occurred. This RFC
+will formally document what a bundle is, anchoring the definition on how bundles support HCA DCP data consumers. This
+RFC will also describe new bundles types that could be supported by the HCA DCP.
 
 ### User Stories
 
-*Share the [User Stories](https://www.mountaingoatsoftware.com/agile/user-stories) motivating this RFC.*
+1. As a data consumer who develops data processing tools (*e.g.* a sequence read aligner), I would like to browse,
+   query, and get primary data from the DSS so that I can use it as testing data during development.
 
-1. As a data consumer who develops data processing tools (*e.g.* a sequence read aligner), I would like to browse, query, and get primary data from the DSS so that I can use it as testing data during development.
-1. As a data consumer who develops analysis tools (*e.g.* a clustering algorithm), I would like to browse, query, and get alignment data from the DSS so that I can use it as testing data during development.
-1. As a data consumer who is a computational biologist, I would like to get expression matrices from the DSS so that I can compare against data I already have.
-1. As a data consumer who is a biologist, I would like to get expression matrices from the DSS so that I can see the expression profile of &lt;my gene of interest&gt; in &lt;my cell type(s) or organ(s) of interest&gt;.
-1. As a data consumer who is a computational biologist, I would like to know what reference data was used to generate processed data in the DSS and I would like to be able to get it for myself so that I can process/analyze my own data with the same references as HCA DCP data.
+1. As a data consumer who develops analysis tools (*e.g.* a clustering algorithm), I would like to browse, query, and
+   get alignment data from the DSS so that I can use it as testing data during development.
 
+1. As a data consumer who is a computational biologist, I would like to get expression matrices from the DSS so that I
+   can compare against data I already have.
 
-## Scientific "guardrails" [optional]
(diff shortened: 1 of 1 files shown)


rfcs/text/0000-bundle-definition.md
```diff
@@ -0,0 +1,156 @@
...
+- **Type**: The name of the bundle type
+- **Definition**: What is and is not in this bundle type
+- **Target consumer**: Who will be consuming this bundle type
+- **Use cases**: Which Users Stories are supported by this bundle type
+
+| Type | Definition | Target consumer | Use cases |
+|:-|:-|:-|:-|
+| Primary bundle | A bundle that contains primary (raw) data files and all related metadata files. | computational biologists; Data Processing Pipelines; Data Browser; Query Service | 1 |
+| Secondary bundle | A bundle that contains secondary (alignment, expression) data files produced by the Data Processing Pipelines, input primary data files, and all related metadata files. | computational biologists; Data Browser; Query Service; Matrix Service | 2, 3, 4(?) |
+| Tertiary bundle* | A bundle that contains tertiary (expression) data files produced by the Matrix Service, input primary and secondary data files, and all related metadata files. | computational biologists; biologists; Data Browser; Query Service | 3, 4 |
```
kislyuk, line 132: I think "tertiary (expression)" is unnecessarily confusing. Tertiary analysis can be any meta-analysis, it should not be constrained to be "expression", which itself is a confusing shorthand for "expression matrix".

I propose renaming:
- "Secondary bundle" -> "Analysis bundle"
- "Tertiary bundle" -> "Expression Matrix bundle"

rfcs/text/0000-bundle-definition.md
```diff
@@ -0,0 +1,156 @@
...
+| Project bundle* | A bundle that contains project-level metadata files. This bundle type does not contain data files or metadata files related to biomaterials, protocols, processes, or files. | computational biologists; biologists; | ? |
+| Resource/Reference bundle* | A bundle that contains reference data files used by the Data Processing Pipelines(?) and all related metadata files. This bundle type does not contain data files or metadata files related to biomaterials, protocols, processes(, or projects?). | computational biologists; Data Processing Pipelines; | 2, 3, 5 |
+| DAPS bundle* | ? | ? | ? |
```
kislyuk: I think there should be a link here to https://github.com/HumanCellAtlas/dcp-community/pull/88, indicating that the term and concept of a DAPS bundle is subject to discussion.

kislyuk (changes requested):

rfcs/text/0000-bundle-definition.md
malloryfreeberg: Good point. I have added a reference.

rfcs/text/0000-bundle-definition.md
malloryfreeberg: I'm ok with renaming secondary and tertiary (in fact, that's the point of this RFC!). If we are renaming those to be nouns, we should also rename the primary bundle, perhaps:

- "Primary bundle" -> "Raw (or Primary) data bundle"
- "Secondary bundle" -> "Analysis bundle"
- "Tertiary bundle" -> "Expression Matrix bundle"

To clarify, "Secondary/Analysis bundles" also contain expression data (gene x cell count matrices) in addition to alignment results, so I wonder if it's confusing to have these results not in the "Expression Matrix bundle".

diekhans: "Raw" doesn't seem very meaningful.  Someone looking at gene expression (most likely the majority of users, might think of unnormalized counts as "raw".

"Matrix" is a type of data structured, so it seems odd to include it in the name.  Kind of like calling sequence experiments "fastaq bundles".   How about "gene expression bundle"

Analysis doesn't have much meaning either.  I was kind of thinking gene expression is the result of analysis.

rfcs/text/0000-bundle-definition.md
```diff
@@ -0,0 +1,156 @@
...
+Describe here...
+
+- Should we call it "Resource" or "Reference" bundle?
+- Who will generate/maintain these? Data Processing Pipelines?
+
+### Acceptance Criteria [optional]
+
+*Acceptance criteria are the conditions that a RFC must satisfy to be accepted by users or other stakeholders.* 
+
+1. Data consumers can make use of data from the DCP without knowing what a bundle is.
```
samanehsan: I see this RFC is still in draft status, but I wanted to note here that this AC doesn't seem to fit in with the rest of the document, which focuses on defining bundle types rather than hiding bundle details from data consumers. Maybe it can be re-worded or removed?

NoopDog:
It may be more useful to tag bundles by the process that created them rather than the type of data they contain. The type of data can be inferred from the process and the process is more descriptive. 

I have a few more notes about this here:

https://github.com/HumanCellAtlas/dcp-community/pull/86#issuecomment-530276016

rfcs/text/0000-bundle-definition.md
kislyuk: Thank you @samanehsan, I agree it's not addressed by the RFC in its current form. I'll remove it.

kislyuk:
@NoopDog how would this tagging look like in practice, and what would a subscription query look like to specifically match such bundles?

diekhans:
It is unclear why there are two RFCs related to bundle (#86 and #93).

They are not clearly delineated into covering different aspects of bundle types.  I believe we would be better off merging these two.

rfcs/text/0000-bundle-definition.md
NoopDog: Seems to me the AC above is a valid constraint on the solution, that it does not further bake in the bundle concept to the DCP interface. Users care about releases, the files in them, and the processes used to create the files. 

We should be careful not to force hundereds/thousands of users to also learn our "bundle types" when we already have metadata describing the processes that creates the files.

NoopDog:
> @NoopDog how would this tagging look like in practice, and what would a subscription query look like to specifically match such bundles?

In practice the least disruptive way would be to add bundle level metadata fields for:
process_fquid - the process instance that created the bundle
protocol_id - the protocol that the process implements - a value form the protocol core schema

Then the user can subscribe to bundles specifying protocol_id along with other refining metadata.

kislyuk:
@NoopDog I appreciate the intent but I think opaque process IDs would still make the system too complex and indirected to be usable.

rfcs/text/0000-bundle-definition.md
kislyuk: @NoopDog That may be true, but the content of this RFC as it stands does not change the status quo in that respect nor pass judgment on how data is presented to the user.

More generally, any advocate for removing bundles from the presentation layer should come up with a concrete, actionable proposal (perhaps in another RFC) for what to replace them with.

rfcs/text/0000-bundle-definition.md
```diff
@@ -0,0 +1,155 @@
...
+
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC number and this pull request in the format:*
+
+`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+
+# HCA DCP Bundle Types and Definitions
+
+## Summary
+
+There are currently no documented definitions for what a "bundle" is in the HCA DCP, including no definitions of what they are, what they contain, or who they support. Lack of clarity causes confusion for data consumers about how the data and metadata in the DCP are structured and organized. Here we propose to formalize a bundle definition, bundle type definitions, target users and consumers, and specific use cases in order to support the needs of data consumers.
```
chmreid: The DRS API documentation provides a "DRS Principles" section that defines their notion of bundles, along with a few other high-level concepts, and may provide a useful example: https://ga4gh.github.io/data-repository-service-schemas/docs/#_drs_api_principles

rfcs/text/0000-bundle-definition.md
```diff
@@ -0,0 +1,155 @@
...
+
+*Recommended format for Shepherds:*
+
+ `[Name](mailto:username@example.com)`
+
+## Motivation
+
+*Describe the user or technical need in detail [with alignment to the DCP roadmap priorities where possible]. Link prior community discussions to demonstrate support for this RFC.*
+
+There are currently no documented definitions for what a "bundle" is in the HCA DCP, including no definitions of what they are, what they contain, or who they support. Lack of clarity causes confusion for data consumers (and DCP developers) about how the data and metadata in the DCP are structured and organized. At the June 2019 DCP F2F meeting in Cambridge, a breakout session was held to discuss "What is a bundle?", and a lot of useful discussion occurred. This RFC will formally document what a bundle is, anchoring the definition on how bundles support HCA DCP data consumers. This RFC will also describe new bundles types that could be supported by the HCA DCP.
```
chmreid:
```suggestion
There are currently no documented definitions for what a "bundle" is in the HCA DCP, including no definitions of what they are, what they contain, or who they support. Lack of clarity causes confusion for data consumers (and DCP developers) about how the data and metadata in the DCP are structured and organized. At the June 2019 DCP F2F meeting in Cambridge, a breakout session was held to discuss "What is a bundle?", and a lot of useful discussion occurred. This RFC will formally document what a bundle is, anchoring the definition on how bundles support HCA DCP data consumers. This RFC will also describe new bundle types that could be supported by the HCA DCP.
```

chmreid: (replaced "new bundles types" with "new bundle types")

kislyuk (approved):

rfcs/text/0000-bundle-definition.md
```diff
@@ -0,0 +1,155 @@
...
+*Describe recommended or mandated review from HCA Science governance to ensure that the RFC addresses the needs of the scientific community.*
+
+## Detailed Design
+
+*Explain the design in sufficient detail such that the implementation and (if appropriate) the interaction with existing DCP software are both reasonably clear.*
+
+### General bundle definition
+
+In the HCA DCP, a bundle is defined as: 
+1. a transaction of files that belong together for scientific reasons - **OR** -
```
TimothyTickle: As we adopt more and more scientific use cases this definition (1) will become less and less useful. With more use cases, files will need to be dynamically aggregated to fulfill specific use cases (as illustrated in the multlibrary RFA). If this is a hard grouping (a transaction of files) then would you group files only by run or only by release or biological network? None of these would be the right answer, but all would be needed groupings for scientific use cases.

rfcs/text/0000-bundle-definition.md
```diff
@@ -0,0 +1,155 @@
...
+#### Table header definitions
+- **Type**: The name of the bundle type
+- **Definition**: What is and is not in this bundle type
+- **Target consumer**: Who will be consuming this bundle type
+- **Use cases**: Which Users Stories are supported by this bundle type
+
+| Type | Definition | Target consumer | Use cases |
+|:-|:-|:-|:-|
+| Primary bundle | A bundle that contains primary (raw) data files and all related metadata files. | computational biologists; Data Processing Pipelines; Data Browser; Query Service | 1 |
+| Secondary bundle | A bundle that contains secondary (alignment, expression) data files produced by the Data Processing Pipelines, input primary data files, and all related metadata files. | computational biologists; Data Browser; Query Service; Matrix Service | 2, 3, 4(?) |
```
TimothyTickle: Pipelines may eventually have to run on "Secondary bundles" as well, we just don't currently.

rfcs/text/0000-bundle-definition.md
TimothyTickle: Gene expression matrices are a result of both pipelines and the matrix service (pipelines make them, matrix service acts on them for users).

rfcs/text/0000-bundle-definition.md
```diff
@@ -0,0 +1,155 @@
...
+| Tertiary bundle* | A bundle that contains tertiary (expression) data files produced by the Matrix Service, input primary and secondary data files, and all related metadata files. | computational biologists; biologists; Data Browser; Query Service | 3, 4 |
+| Project bundle* | A bundle that contains project-level metadata files. This bundle type does not contain data files or metadata files related to biomaterials, protocols, processes, or files. | computational biologists; biologists; | ? |
+| Resource/Reference bundle* | A bundle that contains reference data files used by the Data Processing Pipelines(?) and all related metadata files. This bundle type does not contain data files or metadata files related to biomaterials, protocols, processes(, or projects?). | computational biologists; Data Processing Pipelines; | 2, 3, 5 |
+| DAPS bundle* | Definition discussion ongoing in [PR #88](https://github.com/HumanCellAtlas/dcp-community/pull/88). | ? | ? |
+
```
TimothyTickle, line 136: I am not sure why the above bundles were defined in this manner other than this is de facto, but I am not sure that is the  intent of this RFC, or at least does not feel like a formal understanding of bundles but more of a documentation of what we are using (without comment or systematic definitions).

rfcs/text/0000-bundle-definition.md
```diff
@@ -0,0 +1,155 @@
...
+- Produced by Matrix Service. *Will anyone else be producing expression data files using the output of Data Processing Pipelines as input?*
+- In the first instance will contain expression data files representing one project. In the future, will contain expression data files representing partials projects or multiple projects.
+- 
+
+#### New Project bundle
+
+Describe here...
+
+- Will contain project-level metadata only. 
+- Unclear how this bundle will be used by data consumers.
```
TimothyTickle: How about internal users?

TimothyTickle (changes requested):
I like the way this RFC is heading and the kinds of questions being asked here but I do not think this RFC answers those questions or is even complete.

Happy to talk if you feel I can be helpful.

rfcs/text/0000-bundle-definition.md
```diff
@@ -0,0 +1,155 @@
...
+| Tertiary bundle* | A bundle that contains tertiary (expression) data files produced by the Matrix Service, input primary and secondary data files, and all related metadata files. | computational biologists; biologists; Data Browser; Query Service | 3, 4 |
+| Project bundle* | A bundle that contains project-level metadata files. This bundle type does not contain data files or metadata files related to biomaterials, protocols, processes, or files. | computational biologists; biologists; | ? |
+| Resource/Reference bundle* | A bundle that contains reference data files used by the Data Processing Pipelines(?) and all related metadata files. This bundle type does not contain data files or metadata files related to biomaterials, protocols, processes(, or projects?). | computational biologists; Data Processing Pipelines; | 2, 3, 5 |
+| DAPS bundle* | Definition discussion ongoing in [PR #88](https://github.com/HumanCellAtlas/dcp-community/pull/88). | ? | ? |
+
+> Asterisk (*) indicates bundles types that do not currently exist in the DCP but have been discussed as potential future bundle types to support data consumers.
+
+### New bundle types and their definitions
```
TimothyTickle, line 140: These new bundle types and definitions are not complete but I would love to review them when they are. Feel free to reach out.

rfcs/text/0000-bundle-definition.md
kislyuk: Thanks @chmreid - this is a good reference. Unfortunately, that document only describes some technical details of API design, not the data organization, type hierarchy, or other aspects of bundle operational definitions needed here.

I have imported the contents of #86 and expanded the bundle type definition here with some input from prior DCP discussions/experience.

rfcs/text/0000-bundle-definition.md
kislyuk: Thanks! Fixed.

rfcs/text/0000-bundle-definition.md
kislyuk: @TimothyTickle thanks for pointing this out. I updated and expanded the bundle type definition while merging the contents of #86 into this PR - please take another look.

I disagree that the "transactional unit" definition will become less useful. Operationally, a bundle is the output of an application run. It has to be grouped transactionally for data integrity puproses, and decorated with metadata for discoverability purposes. Whether the grouping is done by release, biological network, etc. is up to the application developers.

rfcs/text/0000-bundle-definition.md
kislyuk: @TimothyTickle I believe that is the intent. I don't think we need to "formally understand bundles". We need to define bundle types and annotate them with enough information for application developers to easily reason about them, so they can develop applications and present data to users with reasonable certainty.

rfcs/text/0000-bundle-definition.md
kislyuk: Thanks, great point. I added "Data Processing Service" to the list above.

rfcs/text/0000-bundle-definition.md
kislyuk: Thanks, great point. Added "and DCP services" to the above.

rfcs/text/0000-bundle-definition.md
```diff
@@ -0,0 +1,248 @@
...
+
+1. As a computational biologist, I would like to use the same references that the HCA DCP uses in their analysis.
+
+## Detailed Design
+
+### What is a bundle?
+
+A bundle is a unit of data organization, roughly equivalent to a folder or directory. In the HCA DCP, a bundle is
+defined as either a transactional grouping of files that belong together for scientific reasons, or a useful association
+between a set of files for utility purposes.
```
xbrianh, line 96: We should consider mentioning that bundles are the units of discoverability in the data-store.

kislyuk:
Please note: I have merged the contents of #86 into this PR; please refer to the discussion in that PR for additional detail and comments addressed from reviewers.

rfcs/text/0000-bundle-definition.md
```diff
@@ -0,0 +1,248 @@
...
+| Secondary Analysis Bundle| Analysis data bundle| `hca/analysis-output; hca-pipeline:snap-atac`               |
+| Reference Resource Bundle| Utility bundle      | `hca/analysis-reference; hca-resource-type:reference-genome`|
+| Expression Matrix Bundle | Analysis data bundle| `hca/analysis-output; hca-resource-type:expression-matrix`  |
+
+The registry would be maintained either as part of this git repository (HumanCellAtlas/dcp-community) or as part of a
+repository chosen by the Data Store team, subject to a pull request-based contribution model.
+
+### Acceptance Criteria [optional]
+
+*Acceptance criteria are the conditions that a RFC must satisfy to be accepted by users or other stakeholders.*
```
xbrianh: Looks like some template strings were included in the RFC text, here and below. Intentional?

xbrianh (approved):

diekhans:
I create issue #114 to get media types turned into a real RFC.  That is good enough to not impact this PR.

rfcs/text/0000-bundle-definition.md
diekhans: agreed with @kislyuk ; transactions are key to data integrity and it is necessary for DCP components to think deeply about data integrity.

kislyuk:
@diekhans I already imported it here: https://github.com/HumanCellAtlas/dcp-community/pull/113

rfcs/text/0000-bundle-definition.md
```diff
@@ -0,0 +1,248 @@
...
+- **Target consumer**: Who will be consuming this bundle type
+
+- **Use cases**: Which Users Stories are supported by this bundle type
+
+| Type | Definition | Target consumer | Use cases |
+|:-|:-|:-|:-|
+| Primary bundle | A bundle that contains primary (raw) data files and all related metadata files. | computational biologists; Data Processing Pipelines; Data Browser; Query Service | 1 |
+| Secondary bundle | A bundle that contains secondary (alignment, expression) data files produced by the Data Processing Pipelines, input primary data files, and all related metadata files. | computational biologists; Data Processing Pipelines; Data Browser; Query Service; Matrix Service | 2, 3, 4(?) |
+| Tertiary bundle* | A bundle that contains tertiary (expression) data files produced by the Matrix Service, input primary and secondary data files, and all related metadata files. | computational biologists; biologists; Data Browser; Query Service | 3, 4 |
+| Project bundle* | A bundle that contains project-level metadata files. This bundle type does not contain data files or metadata files related to biomaterials, protocols, processes, or files. | computational biologists; biologists; | ? |
```
NoopDog, line 133: What is the rational for the "Project Bundle" What use case etc was thais meant to accomplish or what task made easier. Why is this being proposed? Thanks!

rfcs/text/0000-bundle-definition.md
```diff
@@ -0,0 +1,248 @@
...
+| Resource/Reference bundle* | A bundle that contains reference data files used by the Data Processing Pipelines(?) and all related metadata files. This bundle type does not contain data files or metadata files related to biomaterials, protocols, processes(, or projects?). | computational biologists; Data Processing Pipelines; | 2, 3, 5 |
```
NoopDog, line 134: Is it possible to give a bit more detail on the resource bundle? What kind of data goes here? What is the current process for a user to obtain this kind of data?

NoopDog:
As a refinement and extension of this RFC I have created an alternate RFC linked here : [RFC: HCA DCP Application Layer Bundle Types and Definitions ](https://github.com/HumanCellAtlas/dcp-community/pull/119).

The alternate proposal differs from this RFC mainly in that:

1. Type information is added to a new type.json metadata file rather than added to the DSS bundle.json file.

1. Type information is expressed in JSON rather than in [RFC 7231](https://tools.ietf.org/html/rfc7231#section-3.1.1.1) media type syntax.

1. The schema and bundle types are represented as JSON schema in the metadata schema repo and documented on the data portal rather than maintained in a DSS registry.

1. The proposed types are refined by making the process and protocol that created the bundle explicit in the type.json.

Your review/feedback is appreciated.

TimothyTickle (approved):
This looks much better, and the definitions feel less specific to use and a little more generalized and reusable (potentially less over-fit).

rfcs/text/0000-bundle-definition.md
kislyuk: Added an extra sentence for this, thanks.

rfcs/text/0000-bundle-definition.md
kislyuk: Thanks for flagging this. Some template strings are needed in the "Q&A" section for structure. I've removed the rest.

rfcs/text/0000-bundle-definition.md
kislyuk: That is described in the *New Project bundle* section below.

rfcs/text/0000-bundle-definition.md
kislyuk: This was a good gap to address. I added to the *New Resource/Reference bundle* section below:
```
#### New Resource/Reference bundle

+This is a type of bundle that supports reference genomes and other accessory inputs to analysis pipelines. They should
+be represented in the DCP Data Store to achieve completeness and durability of analysis inputs, which in turn is
+necessary for analysis reproducibility.

+- The Data Processing Pipelines team is responsible for generating these.
```

kislyuk: Currently the user would need to look at analysis inputs to recover the URLs for these inputs, and hope that they are still there.

DailyDreaming (approved):

NoopDog:
Pausing to merge with #119 and align with upcoming reproducibility and data citation requirements.

NoopDog:
Obsolete

PR #95, opened 2019-07-30
Hint that the title name should be replaced

mweiden:
Many people get confused about usage here:
* https://github.com/HumanCellAtlas/dcp-community/pull/91#discussion_r308973659
* https://github.com/HumanCellAtlas/dcp-community/pull/90#discussion_r307904536
* ... have observed others

### DIFF
--- a/rfcs/rfc-template.md
+++ b/rfcs/rfc-template.md
@@ -4,7 +4,7 @@
 
 `[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
 
-# RFC Name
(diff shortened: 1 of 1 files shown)


brianraymor (commented):
The charter template is similar. If comprehension is difficult then perhaps it should be:
# Type the name of your RFC here

or the instructions could say:
"Replace *RFC Name* with the name of *your* RFC. Keep it short and descriptive.

brianraymor (changes requested):
See my earlier suggestions for better approaches ...

brianraymor (approved):

PR #98, opened 2019-08-08
Managed Access Data

kozbo:
Update: PM decided to pause this RFC during the 2019-08-22 PM meeting

PR #99, opened 2019-08-09
Distribution of processed data generated from managed access datasets - WIP

lauraclarke:
This RFC proposes a flow of managed access data which enables the DCP to process managed access data and redistribute the summary level data such as matrix files but has the DCP not taking on responsibility for distribution of the managed access data itself, leaving that to 3rd parties

This represents an alternative proposal to [Managed Access Data WIP](https://github.com/HumanCellAtlas/dcp-community/pull/98)


lauraclarke:
Meant to make this as a draft PR, I am going to close and try again

PR #100, opened 2019-08-09
Distribution of processed data generated from managed access datasets

lauraclarke:
This RFC proposes a flow of managed access data which enables the DCP to process managed access data and redistribute the summary level data such as matrix files but has the DCP not taking on responsibility for distribution of the managed access data itself, leaving that to 3rd parties

This represents an alternative proposal to [Managed Access Data WIP](https://github.com/HumanCellAtlas/dcp-community/pull/98)


rfcs/text/0000-distributing-process-results-for-managed-access.md
```diff
@@ -0,0 +1,142 @@
...
+## Scientific "guardrails" 
+
+This plan will be reviewed by members of the [Compliance working group]([https://github.com/HumanCellAtlas/dcp-community/blob/master/charters/Compliance-WG/charter.md](https://github.com/HumanCellAtlas/dcp-community/blob/master/charters/Compliance-WG/charter.md)) who must ensure the proposed design meets their guidelines for responsible handling of managed access genomic data.
+
+## Detailed Design
+
+The proposed data flow for managed access data takes some significant departures from the standard DCP data flow. 
+
+1. The DCP will not redistribute genomic data files which require managed access (fastq and bam).
+2. Genomic data files are still submitted via the ingestion service and HCA data and metadata validation but then brokered to the European Genome-Phenome Archive (EGA) rather than being exported to the Data Store. The EGA has long-standing experience in storing and distributing datasets which need managed rather than open access support which the DCP can leverage.
```
diekhans, line 57: What about genomic data that is already in EGA?

There will be DACs that will require the DCP to be granted access to submit data to the EGA.   How are we going to mange getting approval from a larger number of DACs with different rules?

diekhans (changes requested):
This is a good and very pragmatic proposal and avoids us needed to become distributors of controlled-access data.  For the one case I am familiar with, NCI data, distributing controlled access data requires becoming an NIH Trusted Partner, which is normally only given to institutes with contracts.

I suspect this has long-term value as one mode of operation and would like it to be framed this way, rather than just a near-team bandaid/plaster.

The big question with both this and the related RFC involves dealing with many different DACs with different rules.  We need to be granted access in order to analyze the data. 

Both this and the other controlled access RFC need to investigate what is involved with getting permission from a wide-ranging DACs.  Before deciding on technical solutions, we need to understand the non-technical requirements.  Will we have people specifically responsible for requesting access?  Is this a wrangler task?  What are the application processes like?  How varied are the rules?   

This all needs to be understood in a rigorous manner before thinking about technical approaches.

I would suggest putting both RFCs on hold and doing research and creating a single document that defines the legal and contractual requirements and options for using and managing controlled-access data from a variety of sources.  Gets some PHI experts involved in writing it, the EGA is close by, but we also need to look at NIH rules.

Once we understand the problem, we can have the skills to address it technically.

lauraclarke:
> This is a good and very pragmatic proposal and avoids us needed to become distributors of controlled-access data. For the one case I am familiar with, NCI data, distributing controlled access data requires becoming an NIH Trusted Partner, which is normally only given to institutes with contracts.
> 
> I suspect this has long-term value as one mode of operation and would like it to be framed this way, rather than just a near-team bandaid/plaster.

I changed the title to reflect this isn't a plan which would disappear entirely, even if the DCP did chose to establish the process and technology which allow distribution of managed access data. If I missed other allusions to this being a short term plan do point them out

> The big question with both this and the related RFC involves dealing with many different DACs with different rules. We need to be granted access in order to analyze the data.
> 
> Both this and the other controlled access RFC need to investigate what is involved with getting permission from a wide-ranging DACs. Before deciding on technical solutions, we need to understand the non-technical requirements. Will we have people specifically responsible for requesting access? Is this a wrangler task? What are the application processes like? How varied are the rules?
> 
> This all needs to be understood in a rigorous manner before thinking about technical approaches.
> 
> I would suggest putting both RFCs on hold and doing research and creating a single document that defines the legal and contractual requirements and options for using and managing controlled-access data from a variety of sources. Gets some PHI experts involved in writing it, the EGA is close by, but we also need to look at NIH rules.
> 
> Once we understand the problem, we can have the skills to address it technically.

You are correct that for existing data and new data which goes to the EGA (or dbGAP) before it comes to the DCP, the DCP team getting permission to process the data will be a task which requires careful planning so we can manage the process with as minimal overhead as possible (and that overhead will never by low)

This plan has already been reviewed by representatives from the HCA ethics and regulatory working group and they see value in this approach from a legal/ethical framework perspective.

There are also different axis of difficulty in this space and we need to make sure that we don't move forward on the more straightforward axes because it would be very difficult to solve the more challenging ones.

Those axes as I see them are

**Data under existing dacs and data access agreements vs new DACs which follow HCA ethics and regulatory guidance on access agreements**

A proposal from the HCA ethics working group is that for new data the contributor agreement could contain a clause which gives the DCP team permission to process the data. The first iteraton of this might be with data where the DACs are using the HCA recommended process while we work through the other situation 

**Distribution of processed results vs Distribution of managed access data**

Many if not most data use agreements allow processed non-identifyable results to be distributed by the people who have permission to access the data. Count matrixes fall firmly into this category. The primary goal of this document was to ensure we could move forward giving the majority of the community what they want from these data without holding us back because establishing the framework to allow us to redistribute the genomic data takes longer

**Working with samples/contributors from nations with more liberal laws on genomic data sharing vs nations with strict laws on genomic data sharing**

We won't get data from Indian samples from Indian in the HCA because Indian law doesn't allow genetic information to leave India, the UK, on the other hand, has much more open genomic data sharing regulation which supports responsible sharing of such data and so working with managed access samples which are subject to UK law will be much more straight forward.

This is definitely a place where we should fully leverage our connection to GA4GH as supporting responsible sharing of genomic data is their primary goal and many of the standards they are putting in place recognise the differential challenge in different nations.

This isn't a simple problem but I am hoping that a design pattern like these gives us both near term and long term solutions for the DCP to support the distribution of analysis results from managed access data

mweiden:
Update: PM decided to pause this RFC [during the 2018-08-22 PM meeting](https://docs.google.com/document/d/1SAcKgRg0Acu3baxObiGqrLv1_U-w1ytkOhFiUTG2JHQ/edit#bookmark=id.suguo4xlbcwu)

PR #101, opened 2019-08-09
RFC: DSS Bundle Enumeration

xbrianh:
Proposed addition to the DSS API to enumerate bundles independently of the DSS Elasticsearch metadata index.

Last call for community review: Sept 24

### DIFF
--- a/rfcs/text/0000-dss-bundle-enumeration.md
+++ b/rfcs/text/0000-dss-bundle-enumeration.md
@@ -1,0 +1,43 @@
+### DCP PR:
+
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC number and this pull request in the format:*
+
+`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+
+# RFC: Bundle Enumeration of the DCP Data Store (DSS)
+
+## Summary
+
+This RFC proposes bundle enumeration endpoint(s) for the DSS.
+
+## Author(s)
+
+* [Brian Hannafious](mailto:bhannafi@ucsc.edu)
+
+## Shepherd
+***Leave this blank.** This role is assigned by DCP PM to guide the **Author(s)** through the RFC process.*
+
+*Recommended format for Shepherds:*
+
+ `[Name](mailto:username@example.com)`
+
+## Motivation
+
+Currently, the only means of listing bundles stored in the DSS is the internal Elasticsearch (ES) metadata index, which
+must be kept current with object storage. The DSS should provide bundle enumeration independently of the ES metadata index,
+emphasizing consistency and scalability.
+
+### User Stories
+
+* As a downstream service developer, I would like to enumerate the bundle contents of the DSS so I can create my own
+  index.
+
+* As a downstream service developer, I would like to check if my index contains all the bundles in the DSS.
+
+## Detailed Design
+
+A new bundle enumeration endpoint, `GET /bundles`, will be introduced, taking replica and prefix parameters. These
+parameters will be used to return a paginated listing of bundles directly from object storage. Pagination semantics
+and all other semantics of this route will be in line with the established conventions of the DSS API.
+
+### Unresolved Questions
--- a/rfcs/text/0000-dss-bundle-enumeration.md
+++ b/rfcs/text/0000-dss-bundle-enumeration.md
@@ -36,8 +36,9 @@ emphasizing consistency and scalability.
 
 ## Detailed Design
 
-A new bundle enumeration endpoint, `GET /bundles`, will be introduced, taking replica and prefix parameters. These
-parameters will be used to return a paginated listing of bundles directly from object storage. Pagination semantics
-and all other semantics of this route will be in line with the established conventions of the DSS API.
+A new bundle enumeration endpoint, `GET /bundles`, will be introduced, taking replica and prefix parameters. A listing
+all all bundles having UUIDs beginning with `prefix` will be returned directly from object storage. The prefix
+parameter is intended to facilitate parallelized listings. Pagination semantics and all other semantics of this route
+will be in line with the established conventions of the DSS API.
 
 ### Unresolved Questions
--- a/rfcs/text/0000-dss-bundle-enumeration.md
+++ b/rfcs/text/0000-dss-bundle-enumeration.md
@@ -36,8 +36,8 @@ emphasizing consistency and scalability.
 
 ## Detailed Design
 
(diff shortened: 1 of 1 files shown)


rfcs/text/0000-dss-bundle-enumeration.md
```diff
@@ -0,0 +1,43 @@
...
+### User Stories
+
+* As a downstream service developer, I would like to enumerate the bundle contents of the DSS so I can create my own
+  index.
+
+* As a downstream service developer, I would like to check if my index contains all the bundles in the DSS.
+
+## Detailed Design
+
+A new bundle enumeration endpoint, `GET /bundles`, will be introduced, taking replica and prefix parameters. These
```
Bento007: Can you elaborate on the prefixing parameter?

diekhans:
Since there is no filtering mechanism, other than whatever the prefix does, it seems like the only functional use cases it get a very larger list of all bundles.   If might be nice to state that more explicitly if true.

brianraymor:
@kislyuk - not sure why this is tagged _rfc-community-review_ if it's not actually in community review?

rfcs/text/0000-dss-bundle-enumeration.md
xbrianh: done

rfcs/text/0000-dss-bundle-enumeration.md
```diff
@@ -0,0 +1,44 @@
...
+### User Stories
+
+* As a downstream service developer, I would like to enumerate the bundle contents of the DSS so I can create my own
+  index.
+
+* As a downstream service developer, I would like to check if my index contains all the bundles in the DSS.
+
+## Detailed Design
+
+A new bundle enumeration endpoint, `GET /bundles`, will be introduced, taking replica and prefix parameters. A listing
```
diekhans: It is not clear what the prefix is or its utility.

diekhans: maybe call this "UUID prefix"

rfcs/text/0000-dss-bundle-enumeration.md
```diff
@@ -0,0 +1,44 @@
...
+all all bundles having UUIDs beginning with `prefix` will be returned directly from object storage. The prefix
+parameter is intended to facilitate parallelized listings. Pagination semantics and all other semantics of this route
+will be in line with the established conventions of the DSS API.
+
```
diekhans, line 43: Without some kind of filtering, this seems like a very heavyweight endpoint to use.  A couple of things come to mind:
- Filter by update date/time.  This would allow only obtaining bundles that have changes since the last time the query was run.
- Filter by bundle type.  Well, first we have to have bundle types.

xbrianh: This endpoint is intended for heavyweight use by downstream indexers.

Also, an incremental approach seems preferable: if filtering becomes desirable in the future, it can be added to the endpoint.

diekhans: Agreed that this is easy to add downstream.  Assuming a full dump is what existing users want, then my speculative use is not a real use case ;-)

kislyuk (approved):

rfcs/text/0000-dss-bundle-enumeration.md
kislyuk: @diekhans you raise a good point, but as @xbrianh pointed out, the use case here is an unfiltered bulk pull of all bundle IDs for external indexing. We did look for a way to use a "lightweight" database to do filtering using our established filter language process (JMESpath), but didn't find any suitable database/indexing engine for such a task.

rfcs/text/0000-dss-bundle-enumeration.md
```diff
@@ -0,0 +1,44 @@
...
+* As a downstream service developer, I would like to enumerate the bundle contents of the DSS so I can create my own
+  index.
+
+* As a downstream service developer, I would like to check if my index contains all the bundles in the DSS.
+
+## Detailed Design
+
+A new bundle enumeration endpoint, `GET /bundles`, will be introduced, taking replica and prefix parameters. A listing
+all all bundles having UUIDs beginning with `prefix` will be returned directly from object storage. The prefix
```
brianraymor: _all all_ -> _of all_ ?

rfcs/text/0000-dss-bundle-enumeration.md
```diff
@@ -0,0 +1,44 @@
...
+emphasizing consistency and scalability.
+
+### User Stories
+
+* As a downstream service developer, I would like to enumerate the bundle contents of the DSS so I can create my own
+  index.
+
+* As a downstream service developer, I would like to check if my index contains all the bundles in the DSS.
+
+## Detailed Design
```
brianraymor, line 37: I was expecting something a bit more [swagger-y](https://dss.data.humancellatlas.org/) rather than a narrative description in **Detailed Design** for a API.

rfcs/text/0000-dss-bundle-enumeration.md
xbrianh: fixed

rfcs/text/0000-dss-bundle-enumeration.md
xbrianh: I've updated the text to better reflect the usage of  `prefix` as a "uuid prefix".

rfcs/text/0000-dss-bundle-enumeration.md
xbrianh: @brianraymor It's not clear to me how to address your comment. Perhaps you have something in mind similar to the API endpoint descriptions found in the [Deletion RFC](https://github.com/HumanCellAtlas/dcp-community/blob/0a7cdc63a77581ad891cf6e6bc03bc3046bd8ecc/rfcs/text/0004-dss-deletion-process.md#bundle-deletion-api).

However, those additions will not necessarily improve the clarity and actionability of this document, which defines a simple extension to the DSS API in language that I believe is clear to the developers who will implement it. Do you feel there are technical details missing, or that there is ambiguity that needs to be cleared up?

brianraymor: The Deletion RFC more closely meets my expectations for a RFC as _design document_. Another approach is how Azure defines their [REST APIs](https://docs.microsoft.com/en-us/rest/api/). Any reviewer/developer should be able to read this RFC and understand the DSS API in detail - not just the implementers. This currently reads more like what we called _one pagers_ at Microsoft.

samanehsan (approved):
Looks good! Should this be re-labeled with `oversight-review` now that the `community-review` date has passed?

PR #103, opened 2019-08-09
Data Citation Plan

theathorn:
This RFC outlines a 3-phase plan for providing Data Citation support for the HCA DCP experimental data and associated metadata.

Status: Oversight Review
Last call for oversight review: 4 Oct 2019

*Summary of Review Discussion for Approvers*
There has been general acceptance of the 3 proposed implementation phases with Phase 1 consisting of a stable project URL.
Objections to using an external DOI registration agency (such as Zenodo) have been raised. This matter has been referred to UX for further research, following which a recommendation to the Oversight Committee will be made. 
There was a lengthy discussion on the possible use of Compact Identifiers within the HCA metadata but there is no decision to adopt these at this time.


rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Detailed Design
+
+It is proposed to split the initial implementation into three phases:
+
+### Phase 1
+This is designed to satisfy the minimal set of requirements for User Stories #1 through #4 by providing only "per-project" citations.
+The Data Browser project details page will add a "To cite this project please copy this link" item.
+This "stable URL" will link back to the production site project page using the project's UUID (e.g. https://data.humancellatlas.org/explore/projects/cc95ff89-2e68-4a08-a234-480eca21ce79).
+The URL refers to the “live" view of the project and is therefore subject to additions and updates (e.g. corrections) of data and metadata. However, these are expected to be infrequent and should not affect existing primary data.
```
diekhans: If the version is included in the URL, both the exact version of the project and the live version are available to the user, simply by dropping the version.  This changes it from a loss of information to an interpretation problem,

Using the FQID convention and perhaps modifying the DSS to handle the FQID format rather than using a version parameter could make this easier to implement.

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+If an existing project is deleted and re-ingested then the cited project UUID would become invalid. If such re-ingestion is allowed then a means must be provided to redirect the original "stable URL" to the new version of the project.
```
diekhans: It would be much better to ban reingestion, even if a project has to be updated with a one-off program.

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+Note: Scientists are *already* citing such project based URLs in publications.
+
+See the "Unresolved Questions" section as to whether or not a formal DOI is required for Phase 1.
```
diekhans: Data is commonly cited by accession an occasionally by URL, so I don't think it is a requirement to use DOIs are all.  However, a survey of some relevant journal requirements would help clarify if the URL is sufficient.

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Phase 2
+This is designed to satisfy the data contributor requirements for User Story #5.
+This allows the Data Operations team to make a “Data Release” (i.e. a curated data set) where the data release itself and/or the projects contained within are citable.
+A Data Release must be immutable.
+The Data Browser must provide users with access to each Data Release (i.e. immutable version of data) in addition to the “live/latest” view.
```
diekhans: This is a great point.  Versioning from a user's perspective involves interpretation for a particular task.  For instance, one would often want the exact version of the data with the latest version of the metadata (typos fixed, etc).    RFC in progress to describe this.

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+The stored manifest file could then be used in the HCA CLI to download the exact versions of the data and metadata for that version of the project. It is a requirement that the DCP never deletes older versions of cited data files, except for the special case of retraction of unconsented data.
+
+#### Possible DOI Registration Agencies and open-access repositories that provide the ability to assign DOIs:
+  - [BioStudies](https://www.ebi.ac.uk/biostudies/)
+  - [Zenodo](https://en.wikipedia.org/wiki/Zenodo)
+  - [Figshare](https://en.wikipedia.org/wiki/Figshare)
+  - [Crossref](https://en.wikipedia.org/wiki/Crossref)
+  - HCA DCP assigns its own DOIs by becoming a member of the International DOI Foundation (IDF)
+
+#### DOI Versioning
```
diekhans: Given we have a system where versioning is an integral part of the operation, if seems that DOIs that support versions are a requirement.

diekhans (changes requested):
Overall, really good and to the point.  The only real concern is phase 1 not preserving version.  Given a project could change a lot over time, know exactly what is cited is important to FAIR.

rfcs/text/0000-data-citation-plan.md
theathorn: To include a version in the URL implies support for a versioned view of data as proposed in Phase 2.
Are you saying Phase 1 (live view without versioning view in the Data Browser) is not a viable option?
Even if the Data Browser/Azul only offer a live view it might be feasible to store artifacts (manifest file, medadata tsv, etc) in the DCP or an external (DOI) repository that allowed the user to retrieve those versions of the data via the DCP CLI.

rfcs/text/0000-data-citation-plan.md
theathorn: You meant "ban reingestion"?

rfcs/text/0000-data-citation-plan.md
theathorn: Will bring this up to the UX team.

theathorn: I presume you meant "use DOIs *are* all" to be "Use DOIs *at* all".

rfcs/text/0000-data-citation-plan.md
theathorn: Agreed, and most (all?) do.

mckinsel:
Should we do anything for users who don't get their data via the Data Browser?

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+## Scientific "guardrails" [optional]
+
+*Describe recommended or mandated review from HCA Science governance to ensure that the RFC addresses the needs of the scientific community.*
+
+## Detailed Design
+
+It is proposed to split the initial implementation into three phases:
+
+### Phase 1
+This is designed to satisfy the minimal set of requirements for User Stories #1 through #4 by providing only "per-project" citations.
```
mckinsel: Are you sure a list of links is enough for cc-by attribution?

theathorn: @gabsie Can you comment on this as you raised the original requirement? Is the intent that a data consumer licenses their published work with CC-BY and includes URLs to the DCP in that publication? We aren't currently licensing DCP content with a CC-BY license, so what do we need to do in the DCP to satisfy this requirement?

mweiden:
@mckinsel @theathorn During Phase 1, I guess the assumption is that users will understand how to link the project uuid to the project uuid in the Matrix Service?

theathorn:
> I guess the assumption is that users will understand how to link the project uuid to the project uuid in the Matrix Service?

The citation link (which the Data Browser would supply to the user and may well encode the project_uuid) would get the user to Data Browser project detail page where they could click the "mtx" icon to download the project's matrix.

theathorn:
> Should we do anything for users who don't get their data via the Data Browser?

I think you mean programmatic access via the Matrix Service and/or Query Service? That's currently out-of-scope for this RFC but I'm open to suggestions from Tech Arch as to how this might be achieved in the future.

rfcs/text/0000-data-citation-plan.md
lauraclarke: I would argue that a non-versioned stable URL per project is a viable MVP. We should endeavour to add versioning asap but the first release shouldn't be blocked by not having versioning.

rfcs/text/0000-data-citation-plan.md
lauraclarke: Ingest does plan to automatically submit data to suitable long term resources such as BioStudies, BioSamples and ENA, this gives citable accessions which act as a similar level to DOIs but more granular.

rfcs/text/0000-data-citation-plan.md
lauraclarke: I think given the post prefix text is entirely driven by the service there aren't any who couldn't support versioning, I am not sure if Biostudies can at the moment but it would be a feature we could request

lauraclarke (commented):
> Should we do anything for users who don't get their data via the Data Browser?

@mckinsel given all the info to create the citation is in the metadata hopefully other users won't find it too difficult to assembly, we might want to consider how easy it would be to add a cli or api call that provided citation text for a given project uuid

rfcs/text/0000-data-citation-plan.md
lauraclarke: https://wiki.creativecommons.org/wiki/best_practices_for_attribution

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC number and this pull request in the format:*
+
+`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+
+# HCA DCP Data Citation Plan
+
+## Summary
+
+Data Contributors and Data Consumers need to be able to cite data in the DCP that is referenced in scientific publications.
```
lauraclarke: This seems like the least important use case, if there is already a scientific publication most scientists will reference the publication and not directly link to the DCP

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+
+The creation/update process would perform the following steps:
+  - Ingest creates a new DOI (new project) or a new version of an existing DOI (updated project) when the submission is deemed “complete”.
+  - Upload the corresponding metadata tsv and manifest file for this version of the project to the DOI repository (need to wait for these to be generated).
+  - Upload the matrix output file(s) for this version of the project to the DOI repository.
+  - Update a project description to the DOI repository (e.g. what’s in this version?).
+  - Create a link in the DOI repository back to the project details page in the Data Browser.
+
+### Acceptance Criteria [optional]
+
```
lauraclarke: Please add acceptance criteria for the three phases and success metrics that we can use to see if our solutions actually work for the community

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+### Unresolved Questions
+
+Must a data citation provide an *immutable* view of a project? Or is acceptable for any of the following to undergo additions, updates or deletions over time?
+  - a project's primary data
+  - a project's secondary analysis outputs
+  - a project's expression matrix outputs, generated by the Matrix Service
+  - a project's metadata
+  
+If an *immutable* view of the cited data is a requirement, is this technically feasible before a full implementation of support for versioned files (i.e. the AUDR RFC)?
+
```
lauraclarke: I think supporting citation for the most recent version of a project is the most important goal. In order to reach true fairness we will need to ensure that ultimately citations are tied to specific versions of a project (or data collection) but I believe not putting out a solution before that is possible is letting perfect be the enemy of good.

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+For Phase 1 must a data citation provide a DOI or is a "stable URL" sufficient?
+
```
lauraclarke: I think a stable url is sufficient for MVP. It would be good to understand the cost of DOI assignment so we can work out if it can be done in a suitable time frame for it to be part of the first version of if we are better leaving it to later phases

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+For Phase 2 & 3 does the Data Browser need to provide a view of the cited data on which further faceted searches can be performed?
```
lauraclarke: What is meant by this?

lauraclarke (approved):
This looks like a good plan which I would be happy to see go forward with minor tweaks.

The most important absence from my perspective is clear acceptance criteria for each phase and success metrics which allow us to test if our chosen solution is useful to the community.

rfcs/text/0000-data-citation-plan.md
justincc: I think one could ban re-ingestion for everything except the experiment restructure case, as long as one was prepared for delay in design and implementation. Experiment restructuring is a much more complex case where we don't know if it's realistically possible yet except by re-ingestion. This re-ingestion could preserve the UUID by keeping the existing project metadata, though that may also require AUDR deletion work to delete all the old bundles via ingest before adding new ones.

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+This is designed to satisfy the data consumer requirements for User Story #5.
+This also fully satisfies User Story #2 by allowing a data consumer to cite an arbitrary set of data which spans multiple projects.
+The data consumer must be able to create discrete versions of their selected data and the Data Browser must provide users access to each version.
+The data consumer must be able to share a citable link to any of the versions they have created.
+
+### Implementation notes for DOI support
+A DOI provides a link of the form https://doi.org/xxxx which resolves via the hosting [Registration Agency](https://www.doi.org/registration_agencies.html) or [open-access repository](https://en.wikipedia.org/wiki/Open-access_repository).
+A DOI is the most commonly accepted means for citing documents and data in scientific publications.
+
+It is acceptable for a DOI to resolve to an external website (e.g. BioStudies, Zenodo, etc.) which could then provide a URL to the project in the Data Browser as well as the ability to store point-in-time copies (i.e. versions) of relevant files such as the download manifest, metadata tsv, matrix output file, etc.
```
morrisonnorman: Feedback from the UX team suggests that if someone clicks on a link and expects to go from A to B, but actually goes from A to C to B, this is know as a 'dark pattern' in UX parlance and should be avoided.

When a researcher clicks on a DOI in a paper they expect the DOI resolver to take them to a project page on the HCA-DCP browser where you can download the data.

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+
+#### Possible DOI Registration Agencies and open-access repositories that provide the ability to assign DOIs:
+  - [BioStudies](https://www.ebi.ac.uk/biostudies/)
+  - [Zenodo](https://en.wikipedia.org/wiki/Zenodo)
+  - [Figshare](https://en.wikipedia.org/wiki/Figshare)
+  - [Crossref](https://en.wikipedia.org/wiki/Crossref)
+  - HCA DCP assigns its own DOIs by becoming a member of the International DOI Foundation (IDF)
+
+#### DOI Versioning
+Most DOI repositories appear to provide support for versioning (i.e. multiple versions of a single citation available e.g. via a drop-down selection),
```
morrisonnorman: I'm not sure what is meant by 'DOI repositories'. Figshare or Zenodo are examples of repositories that facilitate citation of the data objects they hold using DOI's.

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+the ability to store accompanying files in the DOI repository (this would be useful for storing versions of manifest and metadata tsv files, etc.),
+and the ability to store URL references (useful for linking back to a page in the Data Browser).
+Examining how both Figshare and Zenodo do DOIs, as the group minting DOIs controls everything after the prefix it is possible to have a main DOI which points to the most recent version and then versioned URLs which point to specific versions.
+
+#### Data Browser
+The Data Browser would contain a reference to the base DOI for each project. Clicking on the DOI link would redirect the user to the (external) DOI repository where the user could see all the versions of that project and download the associated manifest and metadata.tsv files for a specific version.
```
morrisonnorman: As an implementation choice I would be interested for comments by Tech Arch here but this seems to add a lot of overhead and potential for redundancy that would be difficult keep control.

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+Most DOI repositories appear to provide support for versioning (i.e. multiple versions of a single citation available e.g. via a drop-down selection),
+the ability to store accompanying files in the DOI repository (this would be useful for storing versions of manifest and metadata tsv files, etc.),
+and the ability to store URL references (useful for linking back to a page in the Data Browser).
+Examining how both Figshare and Zenodo do DOIs, as the group minting DOIs controls everything after the prefix it is possible to have a main DOI which points to the most recent version and then versioned URLs which point to specific versions.
+
+#### Data Browser
+The Data Browser would contain a reference to the base DOI for each project. Clicking on the DOI link would redirect the user to the (external) DOI repository where the user could see all the versions of that project and download the associated manifest and metadata.tsv files for a specific version. 
+
+#### DOI Creation/Update Process
+Embedding a DOI in the metadata (e.g. biostudies_doi) during ingest would provide per-project citability.
```
morrisonnorman: See my previous comments on dark patterns.

morrisonnorman (changes requested):

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+A Data Release must be immutable.
+The Data Browser must provide users with access to each Data Release (i.e. immutable version of data) in addition to the “live/latest” view.
+The Data Browser must be able to provide users with a means to download the data and metadata associated with a Data Release.
+Since the Matrix Service does not support the ability to process older versions of input files, the output files from the Matrix Service *must* be stored as part of the immutable data set.
+The Data Store "collections" API provides a suitable means for recording the contents of a Data Release.
+
+### Phase 3
+This is designed to satisfy the data consumer requirements for User Story #5.
+This also fully satisfies User Story #2 by allowing a data consumer to cite an arbitrary set of data which spans multiple projects.
+The data consumer must be able to create discrete versions of their selected data and the Data Browser must provide users access to each version.
```
justincc: What do versions mean here. The data releases in phase 2? Or versions created by individual submissions? In the latter case this will need us to increment a project version each time which is a graph of all the file versions after that submission, which is not something we currently do (this begins to sound like a project bundle).

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+This is designed to satisfy the minimal set of requirements for User Stories #1 through #4 by providing only "per-project" citations.
+The Data Browser project details page will add a "To cite this project please copy this link" item.
+This "stable URL" will link back to the production site project page using the project's UUID (e.g. https://data.humancellatlas.org/explore/projects/cc95ff89-2e68-4a08-a234-480eca21ce79).
+The URL refers to the “live" view of the project and is therefore subject to additions and updates (e.g. corrections) of data and metadata. However, these are expected to be infrequent and should not affect existing primary data.
+If an existing project is deleted and re-ingested then the cited project UUID would become invalid. If such re-ingestion is allowed then a means must be provided to redirect the original "stable URL" to the new version of the project.
+Note: Scientists are *already* citing such project based URLs in publications.
+
+See the "Unresolved Questions" section as to whether or not a formal DOI is required for Phase 1.
+
+### Phase 2
```
justincc: So is the data release the point at which each project will get marked as a version, become externally citable and linked to this release?

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Unresolved Questions
+
+Must a data citation provide an *immutable* view of a project? Or is acceptable for any of the following to undergo additions, updates or deletions over time?
+  - a project's primary data
+  - a project's secondary analysis outputs
+  - a project's expression matrix outputs, generated by the Matrix Service
+  - a project's metadata
+  
+If an *immutable* view of the cited data is a requirement, is this technically feasible before a full implementation of support for versioned files (i.e. the AUDR RFC)?
```
justincc: In principle, everything is immutable in the datastore, it's just a matter of finding it. If we had project bundles they could be revved and then point to a completely different set of bundle UUIDs each time. This would require a lot of adaptation by data browser, etc. But then I think you need this even if you have AUDR.

rfcs/text/0000-data-citation-plan.md
diekhans: Given the definition of this phase of citations, I would agree that versions are not essential.  However, I am worried that despite the definitions, project DOIs will be used incorrectly as data citations.  At the very least, we should design for versions, even if we can't implement them initially.

lauraclarke: +1 to design for versioning

People will use DOIs as data citations if we provide them

rfcs/text/0000-data-citation-plan.md
diekhans: I think this depends on the definition of "reingestion".  I take it as "delete everything and start over as if the original data was never there.  While I think the modeling of experiment restructure is best kept basic (e.g. no more than this pile of data use to be this old pile of data), I don't think it should completely lose track of the fact that it is related to the old pile of data.

justincc: Yes, I think this needs a determination of exactly how new data is related to old data. If it doesn't need to be on a bundle by bundle or file by file basis then the problem could become much simpler.

rfcs/text/0000-data-citation-plan.md
theathorn: The cited data (e.g. a Data Release) may contain some files that are older than the latest version. Currently the Data Browser indexes/displays only the latest version of all files. The cited version could be contained in a stored manifest or DSS collection. We would always provide a means to download the cited data via the HCA CLI. This question is asking whether the Data Browser needs to show a "versioned view" of data as well as the latest view and then allow the user to do faceted searches on the versioned view? e.g. The Data Release becomes an additional facet which the users can select (e.g. "latest data" vs "Data Release Q2-2020" vs "Data Release Q3-2020", etc).

rfcs/text/0000-data-citation-plan.md
theathorn: I will revise the RFC to state that Phase 1 will be "a non-versioned stable URL per project" (no DOI).

rfcs/text/0000-data-citation-plan.md
theathorn: Should I rephrase to state that authors of publications need to provide a citation to data in the DCP?

rfcs/text/0000-data-citation-plan.md
theathorn: Agreed.

rfcs/text/0000-data-citation-plan.md
theathorn: I will replace "versions" with "collections". e.g. a user could put an arbitrary selection of files into something like a shopping cart that is persistent. They can retrieve that exact set of data at  any time and share it with others. I'm not envisioning any additional support would be necessary from other components.

rfcs/text/0000-data-citation-plan.md
theathorn: Broadly, yes. Details will need to be worked out when we define what exactly constitutes a Data Release with the Data Operations team. But I'm proposing that, at a minimum, each project within a Data Release would have it's own citation.

rfcs/text/0000-data-citation-plan.md
theathorn: @lauraclarke Where is this CC-BY attribution going and who is creating it? Is a publication author putting the CC-BY license in their publication and linking to the DCP project page? Or is the DCP going to attach CC-BY licenses automatically to each project or to the site as a whole?

rfcs/text/0000-data-citation-plan.md
theathorn: I think this implies that the DCP must obtain it's own DOI prefix (e.g https://www.crossref.org/get-started/content-registration/) and would exclude solutions such as Zenodo (and Biostudies?). The potential advantage of using an external repository that provides versioning is that we could store manifests there for multiple versions of a project instead of being forced to implement those views within the DCP itself. i.e. The user would always see the latest view of a project in the Data Browser but the (DOI) citation link would take them to an external DOI repository where they could view older versions of the project and where manifests for those older versions would be stored, thus enabling the user to download the older manifest from the external repository and use the HCA CLI to retrieve the specified versions of files. I recall seeing at least one other genomic data site that use Zenodo (can't remember which one). @lauraclarke I need some guidance from PLs or Oversight as to whether the "the DOI resolver takes a user directly to a project page on the HCA-DCP browser" is a hard requirement.

rfcs/text/0000-data-citation-plan.md
theathorn: Yes "repositories that facilitate citation of the data objects they hold using DOI's.". The may have some advantages for the DCP (see above conversation). Of course we can also embed link(s) back to the DCP in any content (web page) we store there.

rfcs/text/0000-data-citation-plan.md
lauraclarke: I think the important functionality is that an a user has someway (ideally both programatic and web page) to discover that an old UUID is no longer in use and some sort of pointer to where to do to get the same data, MVP could be old UUID to project id ideally with some sort of text explaining what sort of changed happened

rfcs/text/0000-data-citation-plan.md
lauraclarke: So the cc-by attribution should likely be at the project level the same way as DOIs will be 

If you look at figshare, https://figshare.com/articles/Malignant_Cancer_Cell_Nucleus/9751670
Zenodo https://zenodo.org/record/3363060#.XWoe8ZNKjOQ they both put it on the individual study/project pages

Please note this cc-by license is different from if we as the DCP chose to license the static content of our browser, that sounds like something we should discuss but not here

lauraclarke: I actually think the two sentences you have about data consumers and data contributors works as a stand alone summary and covers the major use cases for us providing stable urls and dois. I think the in scientific publications is implied and doesn't need to be explicitly stated as that it the majority use case for needing a citable record

We can and should ensure that if data has a publication associated with it that users can discover that publication but that is a separate issue

lauraclarke: I guess you could add a sentence like this at the end

"Providing citable records for the DCP will enable all contributors and consumers to reference what data stored in the DCP they used in their scientific publications"

rfcs/text/0000-data-citation-plan.md
lauraclarke: @theathorn @morrisonnorman we should work with UX to come up with pros and cons of using an external authority vs setting up the DCP as a DOI assigning entity and then if that effort doesn't give us a clear answer we should ask Oversight for their view

rfcs/text/0000-data-citation-plan.md
lauraclarke: @justincc just to be clear, data releases being citable doesn't remove the need for individual projects to be citable and if I have understood @theathorn's proposals correctly these comments are to ensure the design work enables data releases to be citable once we understand how data releases are produced

rfcs/text/0000-data-citation-plan.md
theathorn: Updated the Summary. I didn't alter the CC-BY User Story, assuming that scientific authors will include such a license in their publications. Should we be including a CC-BY license on each Project Detail page of the Data Browser? i.e. Are we saying the data files for each DCP project are licensed under CC-BY?

rfcs/text/0000-data-citation-plan.md
theathorn: Updated Phase 1 to include a redirection facility for deleted and re-ingested projects.

rfcs/text/0000-data-citation-plan.md
theathorn: Is "project.biostudies_accessions" currently being populated or will it be in the future? If populated I believe it will already show up on the Project Detail page.

I removed the requirement for a formal DOI in Phase 1.

rfcs/text/0000-data-citation-plan.md
theathorn: Updated Phase 1 to include stable non-versioned URLs  and Phase 2 to include support for versioning.

rfcs/text/0000-data-citation-plan.md
theathorn: Clarified that versioned projects are citable for Phase 2.

rfcs/text/0000-data-citation-plan.md
theathorn: Phase 3 updated.

rfcs/text/0000-data-citation-plan.md
theathorn: Added to Unresolved Questions.

rfcs/text/0000-data-citation-plan.md
theathorn: Clarified definition of open access repositories that provide DOI minting services.

rfcs/text/0000-data-citation-plan.md
lauraclarke: We don't currently populate it automatically but certain do plan to. If it would help with getting data citation sorted we could talk with @justincc and @morrisonnorman about if that work can be prioritized

rfcs/text/0000-data-citation-plan.md
theathorn: Added acceptance criteria for each phase - please review.

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Phase 1
+This is designed to satisfy the minimal set of requirements for User Stories #1 through #4 by providing only "per-project" citations.
+The Data Browser project details page will add a "To cite this project please copy this link" item.
+This "stable non-versioned project URL" will link back to the production site project page using the project's UUID (e.g. https://data.humancellatlas.org/explore/projects/cc95ff89-2e68-4a08-a234-480eca21ce79).
+The URL refers to the “live" view of the project and is therefore subject to additions and updates (e.g. corrections) of data and metadata. However, these are expected to be infrequent and should not affect existing primary data.
+If an existing project is deleted and re-ingested then the cited project UUID would become invalid. If such re-ingestion occurs then a means will be provided to redirect the original "stable project URL" to the new version of the project (e.g. by providing a landing page which states something similar to "This project has been updated with corrected data and/or analyses. For the current version of this project click *here*", where *here* is a link to the new project page).
```
justincc: In order to get the Meyer dataset to complete status in a short timeframe [we plan to re-ingest but preserve the existing project UUID](https://docs.google.com/document/d/1O8QHJmMhImsx6OKAPAos3OKPxE9q9UniHHsPbCuAeuo/edit#heading=h.5iqo8bejsuxm). Once this is done we don't anticipate further situations where re-ingestion will lose the original project UUID.

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,153 @@
...
+  - The Data Operations team is able to create an immutable citation reference for each project in a Data Release
+  - The citation reference consists of a versioned DOI
+  - The Data Operations team can update the version of the project's DOI in a later Data Release
+  - The Data Browser provides a means of downloading the data associated with a specific version of a project from a Data Release
+#### Phase 3 (in addition to Phase 1 and 2 criteria)
+  - An authorized user may create a citation for an arbitrary set of versioned data selected via the Data Browser
+  - An authorized user may update a previously versioned set of data to a new version and update the version of the corresponding citation
+   - An authorized user may share a citation which they have created with others users who can then access the specific version of the data referenced by the citation
+
+### Unresolved Questions
```
justincc: Any plan to provide stable URLs for citation within a project (e.g. to within a particular bundle within it)? If not could you state that explicitly since it has major implementation requirements.

theathorn: Updated to address your concern.

rfcs/text/0000-data-citation-plan.md
theathorn: @lauraclarke Can you answer the question in my previous comment? I think I may be misunderstanding something here.

rfcs/text/0000-data-citation-plan.md
theathorn: @lauraclarke Are the acceptance criteria, err, acceptable?

rfcs/text/0000-data-citation-plan.md
theathorn: As agreed, Phase 1 has "stable URLs" only and DOIs are introduced in Phase 2.

rfcs/text/0000-data-citation-plan.md
theathorn: Added clarifying text - OK to resolve this?

rfcs/text/0000-data-citation-plan.md
theathorn: @morrisonnorman If your concerns have been addressed, may I resolve this conversation?

rfcs/text/0000-data-citation-plan.md
theathorn: @diekhans If your concerns have been addressed, may I resolve this conversation?

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,151 @@
...
+Currently the Data Browser provides only a view of the latest data and metadata. It may be desirable to add another "facet" such as "Release Version" which would enable the user to view a particular point-in-time Data Release and then perform further faceted searches within that view.
+
+Can a DOI for the DCP resolve to an external website? The author(s) will work with the UX team to come up with pros and cons of using an external authority vs setting up the DCP as a DOI assigning entity and then, if that effort doesn't give us a clear answer, we will ask the Oversight Committee for their view.
+
+### Drawbacks and Limitations [optional]
+
+*Why should this RFC **not** be implemented?*
+
+### Prior Art [optional]
+
```
hewgreen: Some emphasis on the use cases of programatic consumers, who will likely be your highest frequency users, would be good. Through the lens of current best practice you have only indirectly (through BioStudies) mentioned compact identifiers, but these are critical and would help at all phases of delivery. Consideration of best practice should go into the prior art section.

DOI and URL have both been mentioned. These are relevant to publishers and a user with a keyboard. But compact identifiers are most frequently used to consume, map, link and share data and to my mind represent best practice. They are also the most useful way to discuss a project via word of mouth or via slack which is essential for developers, wranglers and data ops. Saying 'prof x's latest organ y dataset' is not scalable and uuids are not human readable or memorable, which increases the likelihood of miscommunication (missing use case). Within the EBI, these identifiers are used to make data interoperable but all major biological data institutions also embrace them. As a specific example the successful INSDC (sharing genomic data between NCBI, EBI and DDBJ) works by mapping these identifiers.

You will have come across identifiers that look like this in our metadata SRAxxxx, GEOxxxx, PMCxxxx, PRJxxxx... We use them to reference sources of data. This is the same use case as your downstream users. This single short identifier can be resolved similar to a DOI through identifiers.org, communicated in metadata easily and thanks to the resource specific prefix bakes in the source of the data. Bioinformaticians are used to dealing with these identifiers in this format.

As it stands uuids are not HCA specific. Thankfully no-one else (in our realm) seems to use them so we (at SCEA) guess these identifiers refer to a HCA dataset. It would be preferable to have an ID HCAxxxx to confirm this.

At the moment our metadata would look like this:
`"data source" : ["ERP114453", "PRJEB31843", "c4077b3c-5c98-4d26-a614-246d12c2e5d7"]`
Spot the odd one out? How do I resolve this if anyone other than the HCA starts to use uuids?

Other resources also support these identifiers for different granularities of the data. For example ENA provide prefixes to describe studies, experiments, samples and runs. This is very useful for downstream users like us who want to split and splice datasets but still need to refer to the provenance of the data.

So compact identifiers give you a familiar way to communicate granularity and source of data while easing programatic or manual access. These identifiers are often used in the literature to refer to data rather than DOIs. If the plan is for BioStudies to mint these identifiers for HCA projects, I would ask the following:

- This functionality is added ASAP. Once a referencing scheme is established it is difficult and costly to change. If uuids/URLs are referenced in upcoming papers the trend will quickly stick.
- This identifier is in the metadata before a dataset is released so that it can always be relied upon downstream.
- The prefix maintains provenance so it is clear the dataset is HCA
- HCA is able to provide references to more granular components of a project (sample, assay etc) in the future.

briandoconnor (approved):
we reviewed in person, lgtm

rfcs/text/0000-data-citation-plan.md
theathorn: Added to Unresolved Questions.

rfcs/text/0000-data-citation-plan.md
theathorn: The project metadata does contain compact identifiers in the form of "project.insdc_project_accessions" (e.g. SRP078321), "project.insdc_study_accessions" (e.g. PRJNA328774), "project.array_express_accessions" (e.g. E-GEOD-84133), "project.geo_series_accessions" (e.g. GSE84133), and "project.biostudies_accessions".
@lauraclarke Are you on board with adding HCA-specific accessions (e.g. HCAxxxxx) to the project metadata and possibly to other objects (e.g. samples, assays)? Is biostudies_accessions ever populated (I couldn't find an example)? Noting that the INSDC/GEO/ArrayExpress accessions are not always populated for a project and don't provide the direct reference to data in the DCP that would be provided by an HCA compact identifier.

@hewgreen for programmatic consumers, are you expecting additions to an existing API or proposing a new service, and what functions would be needed?

diekhans:
I have absolutely no problem with phase 1 as long as we are making it clear we are not citing specific sets of data.

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,148 @@
...
+2. As a data consumer (e.g. researcher with a keyboard), I want to be able to view and share a unique citation identifier so that a reader of my manuscript can obtain the data needed to reproduce my results. Anyone can use the citation identifier to view and download all the original cited data and metadata files for a project from the DCP.
+3. As a data consumer, I need a simple way to reference a project in the DCP so that I can fulfill the requirements of a Creative Commons attribution license (CC-BY).
+4. As a data contributor or consumer, I need a way to use the citation identifier to access the output produced by the DCP Matrix Service for the data being cited.
+5. As a data contributor or consumer, I want to be able to update my project and provide versioned views of the data being cited.
+
```
lauraclarke: Some of the design and acceptance criteria are framed around the ideas of data releases created by the data operations team or around data consumers creating arbitary sets of data to support their work.

Neither of these concepts are reflected in the user stories

Also, why would a data consumer need to be able to update a project, someone who is solely a consumer won't own any projects so won't have the rights to update anything

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,148 @@
...
+  - In which phase, if any, is support for a formal digital object identifier ([DOI](https://en.wikipedia.org/wiki/Digital_object_identifier)) required?
+
+### User Stories
+
+1. As a data contributor (e.g. researcher with a pipette), it is essential to have a unique way for my project (consisting of primary and secondary analysis data files and associated metadata files) to be identified so that others can properly cite my work when using my data, and that this citation identifier is available in the DCP Data Portal. Note: This will discourage but not prevent others from using my data without attribution.
+2. As a data consumer (e.g. researcher with a keyboard), I want to be able to view and share a unique citation identifier so that a reader of my manuscript can obtain the data needed to reproduce my results. Anyone can use the citation identifier to view and download all the original cited data and metadata files for a project from the DCP.
+3. As a data consumer, I need a simple way to reference a project in the DCP so that I can fulfill the requirements of a Creative Commons attribution license (CC-BY).
```
lauraclarke: This is a side task rather than something relevant to the RFC but I don't think we make the CC-BY license very obvious for our data. Most people won't know that the data is licensed like that

rfcs/text/0000-data-citation-plan.md
lauraclarke: So I think we need to review how we support people understanding that the data is licensed using cc-by

I think the citation widget gives people a way to meet cc-by attribution needs

They will need to be able to add attribution to whenever they reuse something, at the project level seems a good starting point, Ultimately we might want a way for someone to give us any identifier and getting an appropriate attribution text for that identifier in our system

lauraclarke: A paper which might be useful in considering solutions https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0213090

lauraclarke: And a blog post https://nlmdirector.nlm.nih.gov/2019/06/11/socio-legal-barriers-to-data-reuse/

rfcs/text/0000-data-citation-plan.md
lauraclarke: Thanks for the acceptance criteria, these look good but the phase 3 criteria seem disconnected from the user stories as they are written now. The user stories don't mention a data release

rfcs/text/0000-data-citation-plan.md
tburdett: I hate to cite a paper I am myself an author on, but it is worth reading this article before deciding whether or not to create new accessions: https://journals.plos.org/plosbiology/article?id=10.1371/journal.pbio.2001414. This article also references many other useful papers on identifying resources in the life sciences.

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,148 @@
...
+  - Update a project description to the DOI repository (e.g. what’s in this version?).
+  - Create a link in the DOI repository back to the project details page in the Data Browser.
+
+### Acceptance Criteria [optional]
+
+#### Phase 1
+  - A citation link is provided in the Data Browser for each project
+  - The citation link remains valid for the lifetime of the DCP
+  - The citation link is a URL which resolves to the latest version of the project details page for the specified project
+  - The citation URL is used in publications by scientists to cite DCP project data
```
lauraclarke: This is a great statement but it is a success metric (which we should track) rather than an acceptance criterion. This won't be true when the feature is released but we should make sure we measure how frequently people cite us so we can understand the success of the feature

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Detailed Design
+
+It is proposed to split the initial implementation into three phases:
+
+### Phase 1
+This is designed to satisfy the minimal set of requirements for User Stories #1 through #4 by providing only "per-project" citations.
+The Data Browser project details page will add a "To cite this project please copy this link" item.
+This "stable non-versioned project URL" will link back to the production site project page using the project's UUID (e.g. https://data.humancellatlas.org/explore/projects/cc95ff89-2e68-4a08-a234-480eca21ce79).
+The URL refers to the “live" view of the project and is therefore subject to additions and updates (e.g. corrections) of data and metadata. However, these are expected to be infrequent and should not affect existing primary data.
```
lauraclarke: I think the statement "should not affect existing primary data" is overstated.

I think updates will rarely affect primary data (apart from adding new primary data to an existing project) but I don't think it will never affect primary data

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,148 @@
...
+A stable URL applies only to a project as a whole; there is no facility to provide separate citations for individual bundles or files within a project.
+If an existing project is deleted and re-ingested then the cited project UUID would become invalid. If such re-ingestion occurs then a means will be provided to redirect the original "stable project URL" to the new version of the project (e.g. by providing a landing page which states something similar to "This project has been updated with corrected data and/or analyses. For the current version of this project click *here*", where *here* is a link to the new project page).
+Note: Scientists are *already* citing such project based URLs in publications.
+A formal DOI is *not* required for Phase 1.
+
+### Phase 2
+This is designed to satisfy the data contributor requirements for User Story #5.
```
lauraclarke: The framing of user story 5 doesn't provide a strong connection to this phase description

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,148 @@
...
+This allows the Data Operations team to make a “Data Release” (i.e. a curated data set) in which the specific versions of projects within the Data Release itself are citable.
+A Data Release must be immutable.
+The Data Browser must provide users with access to each Data Release (i.e. immutable version of data) in addition to the “live/latest” view.
+The Data Browser must be able to provide users with a means to download the data and metadata associated with a Data Release.
+Since the Matrix Service does not support the ability to process older versions of input files, the per-project output files from the Matrix Service *must* be stored as part of the immutable data set.
```
lauraclarke: Do we know if this is an immovable characteristic of the matrix service? Never supporting older files seems bold decision this early in the project given there must be risk of these files changing?

That said, having all processed outputs stored associated with an immutable release seems like a good idea even if the matrix service could still produce them

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,148 @@
...
+5. As a data contributor or consumer, I want to be able to update my project and provide versioned views of the data being cited.
+
+## Scientific "guardrails" [optional]
+
+*Describe recommended or mandated review from HCA Science governance to ensure that the RFC addresses the needs of the scientific community.*
+
+## Detailed Design
+
+It is proposed to split the initial implementation into three phases:
+
```
lauraclarke: The phase description are written assuming someone already can see what the main goal of the phase is from reading the above text. For understanability it would be nice to give the three phases tldr style titles or summary sentences for the main goal otherwise it leads to a lot of flitting backwards and forwards to understand what is being references

If I have understood correctly I would suggest

Phase 1 - stable non versioned citable URLs
Phase 2 - versioned urls and support for data release citation
Phase 3 - Enabling citation of user defined data collections

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,148 @@
...
+#### Data Browser support for DOIs
+The Data Browser could contain a reference to the base DOI for each project. Clicking on the DOI link would then redirect the user to the (external) DOI repository where the user could see all the versions of that project and download the associated manifest and metadata.tsv files for a specific version. 
+
+#### DOI Creation/Update Process
+Embedding a DOI in the metadata (e.g. biostudies_doi) during ingest would provide per-project citability.
+Provided the DOI repository supports versioning the metadata DOI field could be updated automatically by Ingest to a new version whenever a project update is processed.
+Using a DOI that links to an external repository that can store a manifest for each version of a project may be the simplest way to provide access to versions of the data for a project.
+
+The creation/update process would perform the following steps:
+  - Ingest creates a new DOI (new project) or a new version of an existing DOI (updated project) when the submission is deemed “complete”.
```
lauraclarke: The word "complete" here is unclear. What is meant by complete? There are many ways to define that, more specific details here would be useful

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,148 @@
...
+
+#### Phase 1
+  - A citation link is provided in the Data Browser for each project
+  - The citation link remains valid for the lifetime of the DCP
+  - The citation link is a URL which resolves to the latest version of the project details page for the specified project
+  - The citation URL is used in publications by scientists to cite DCP project data
+#### Phase 2 (in addition to Phase 1 criteria)
+  - The Data Operations team is able to create an immutable citation reference for each project in a Data Release
+  - The citation reference consists of a versioned DOI
+  - The Data Operations team can update the version of the project's DOI in a later Data Release
```
lauraclarke: Do you mean project or release here? why would the data operations team need to update an individual project's DOI?

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,148 @@
...
+Must a data citation that provides a DOI provide an *immutable* view of all the data and metadata associated with a project? Or is acceptable for any of the following to undergo additions, updates or deletions over time?
+  - a project's primary data
+  - a project's secondary analysis outputs
+  - a project's expression matrix outputs, generated by the Matrix Service
+  - a project's metadata
+  
+For Phase 2 & 3 does the Data Browser need to provide a view of the cited data on which further faceted searches can be performed?
+Currently the Data Browser provides only a view of the latest data and metadata. It may be desirable to add another "facet" such as "Release Version" which would enable the user to view a particular point-in-time Data Release and then perform further faceted searches within that view.
+
+Can a DOI for the DCP resolve to an external website? The author(s) will work with the UX team to come up with pros and cons of using an external authority vs setting up the DCP as a DOI assigning entity and then, if that effort doesn't give us a clear answer, we will ask the Oversight Committee for their view.
```
lauraclarke: I would rephrase this

"If the results of the UX and Browser team research does not provide sufficient clarity we will present our prefered option and evidence to the Oversight Committee and ask for their feedback"

rfcs/text/0000-data-citation-plan.md
lauraclarke: I agree with @morrisonnorman we need to think through the implementation plan here to minimize risk of drift and avoid sending users to other services when we don't have to

rfcs/text/0000-data-citation-plan.md
lauraclarke: @tburdett beat me to it but to agree with him. I think we need to think through the consequences of adding another namespace to the identifiers within bioinformatics. There are upsides to do it but there are also costs and we need to ensure we understand them before we embark down that route.

rfcs/text/0000-data-citation-plan.md
morrisonnorman: @theathorn I still have some concerns over the sustainability of using a 3rd party site for storing manifests and minting DOI's. However, I think that is a different concern. Conversation resolved.

morrisonnorman (approved):

rfcs/text/0000-data-citation-plan.md
diekhans: I am correct in that phase 3 would be the fast-path for a producer who wants to cite their own data?

rfcs/text/0000-data-citation-plan.md
hewgreen: @theathorn As you rightfully point out the compact identifiers you refer to are to secondary resources. They do not reflect HCA data (which can be novel, analysed or truncated). They are not always present and cannot be relied upon. Therefore, these do not address the requirements I laid out.

@tburdett the paper outlines the considerations perfectly. One interesting idea is human readable labels. Maybe these could solve some issues (not user provided titles). But overall I think it suggests we should accession in some cases and think carefully about identifier design. UUIDs maybe not being great. I recommend the paper to others. My interpretation of the 'lessons' are as follows:

**1. Credit any derived content using its original identifier**
We should create IDs for HCA's new data and new analysis because it's novel stuff. Source IDs are important.

**2. Help local IDs travel well: Document prefix and patterns**
Prefixes are super important. aka a uuid alone isn't useful.

**3. Opt for simple, durable web resolution**
We should get a resolution provider. Why I mention identifiers.org but DOI sort of works for this although you have to follow them to work out what they are. Our URL prefixes are already nice and simple but the local ID's are long, cumbersome, human unrepeatable.

**4. Avoid embedding meaning or relying on it for uniqueness:**
We don't want to pack meaning into our IDs. Fair point. I don't think anyone want's this but we should be careful not to.

**5. Design new identifiers for diverse uses by others**
The ID's need to be designed to be useful. UUIDs cannot be described as user friendly. We could design something much nicer.

**8. Make URIs clear and findable & 10. Reference and display responsibly**
Our prefixes aren't in the metadata as passed from the DSS. Should they be if they are the only way downstream consumers know a uuid belongs to the HCA?

The following lessons I think we have learned and are working towards already:
**6. Implement a version-management policy**
**7. Do not reassign or delete identifiers**
**9. Document the identifiers you issue and use**

rfcs/text/0000-data-citation-plan.md
theathorn: Raised https://app.zenhub.com/workspaces/orange-5d680d7e3eeb5f1bbdf5668f/issues/humancellatlas/data-browser/902

rfcs/text/0000-data-citation-plan.md
theathorn: Changed as suggested.

rfcs/text/0000-data-citation-plan.md
theathorn: Done.

rfcs/text/0000-data-citation-plan.md
theathorn: Added summary titles.
Phase 3 would enable producers and consumers to cite an arbitrary set of versioned data.

rfcs/text/0000-data-citation-plan.md
theathorn: Clarified Matrix output requirements.

rfcs/text/0000-data-citation-plan.md
theathorn: Updated User Stories #5 (Data Operations team) and #6 (User collections).
An (authorized) data consumer will be able to define their own "research project" as an arbitrary  collection of data they have selected (e.g. "brain data from all projects"). They create a citation for that data. Later on more brain data is added to the DCP and the data consumer creates a new version of the the citation for their research project,

rfcs/text/0000-data-citation-plan.md
theathorn: Removed "used in publications" from acceptance criteria.

rfcs/text/0000-data-citation-plan.md
theathorn: Attempted to clarify - OK now?

rfcs/text/0000-data-citation-plan.md
theathorn: Added Data Release User Story.

rfcs/text/0000-data-citation-plan.md
theathorn: Added Data Release User Story.

rfcs/text/0000-data-citation-plan.md
theathorn: Clarified to explain that a project may be subject to updates between Data Releases.

lauraclarke:
> > Should we do anything for users who don't get their data via the Data Browser?
> 
> I think you mean programmatic access via the Matrix Service and/or Query Service? That's currently out-of-scope for this RFC but I'm open to suggestions from Tech Arch as to how this might be achieved in the future.

@theathorn Maybe a feature request for the HCA cli would be a get citation which would return the same text as the data portal widget does, at least make it easy for programmatic users to get the info without having to browse the website and find what they are looking for.

rfcs/text/0000-data-citation-plan.md
jahilton: I have proposed a F2F session on Accessions, fyi, with a proposal to put them in use on HCA data.

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,131 @@
...
+2. As a data consumer (e.g. researcher with a keyboard), I want to be able to view and share a unique citation identifier so that a reader of my manuscript can obtain the data needed to reproduce my results. Anyone can use the citation identifier to view and download all the original cited data and metadata files for a project from the DCP.
+3. As a data consumer, I need a simple way to reference a project in the DCP so that I can fulfill the requirements of a Creative Commons attribution license (CC-BY).
+4. As a data contributor or consumer, I need a way to use the citation identifier to access the output produced by the DCP Matrix Service for the data being cited.
+5. As a member of the Data Operations team I want to be able to create citations for each of the projects that I include in a Data Release (which is defined as a point-in-time selection of specific verions of data from the DCP).
```
jahilton: I think one should be able to create a citation for each project, regardless of inclusion of a Data Distribution. So maybe this is better...
>As a data consumer, I am going to publish research based on HCA data and I want to create a citation that references the specific version of each project that I used.

jahilton: ...and just read User Story 6, so I think that covers this. I change my suggestion to remove 5, fix the typo in 6 ('contributor') and remove 'or data consumer' in 6 (a data consumer updating a project is not on the table)

jahilton: ...reading further still, I see that a 'research project' in 6 actually seems to refer to a 'data consumer-curated collection of HCA data'
I think my suggested user story above covers both cases of DataOps cutting a Data Distribution and a consumer making their own collection. If each version of each project can be referenced, then any collection is just a list of those references, right?
Unless there is going to be a difference between a collection made by someone DCP internal vs external

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,131 @@
...
+A stable URL applies only to a project as a whole; there is no facility to provide separate citations for individual bundles or files within a project.
+If an existing project is deleted and re-ingested then the cited project UUID would become invalid. If such re-ingestion occurs then a means will be provided to redirect the original "stable project URL" to the new version of the project (e.g. by providing a landing page which states something similar to "This project has been updated with corrected data and/or analyses. For the current version of this project click *here*", where *here* is a link to the new project page).
+Note: Scientists are *already* citing such project based URLs in publications.
+A formal DOI is *not* required for Phase 1.
+
+### Phase 2 - Versioned data release citations
+This is designed to satisfy the requirements for User Story #5.
+This allows the Data Operations team to make a “Data Release” (i.e. a curated data set) in which the specific versions of projects within the Data Release itself are citable.
```
jahilton: Data Release --> Data Distribution

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,131 @@
...
+
+#### Phase 1
+  - A citation link is provided in the Data Browser for each project
+  - The citation link remains valid for the lifetime of the DCP
+  - The citation link is a URL which resolves to the latest version of the project details page for the specified project
+#### Phase 2 (in addition to Phase 1 criteria)
+  - The Data Operations team is able to create an immutable citation reference for each project in a Data Release
+  - The citation reference consists of a versioned DOI
+  - The Data Operations team can update the version of a project's DOI in a later Data Release, for example when updates have been made to that project's data since the previous Data Release.
+  - The Data Browser provides a means of downloading the data associated with a specific version of a project from a Data Release
```
jahilton: This seems beyond the scope of Data Citation and into the realm of Data Distribution access. Would you consider relieving this criteria from the current proposal? Or am I missing the connection?

rfcs/text/0000-data-citation-plan.md
```diff
@@ -0,0 +1,131 @@
...
+#### Phase 3 (in addition to Phase 1 and 2 criteria)
+  - An authorized user may create a citation for an arbitrary set of versioned data selected via the Data Browser
```
jahilton: I think defining 'authorized user' would be helpful

rfcs/text/0000-data-citation-plan.md
theathorn: Done.

rfcs/text/0000-data-citation-plan.md
theathorn: Added definition to User Story #6.

rfcs/text/0000-data-citation-plan.md
theathorn: Clarified to state that users must be able to download the cited data that makes up a Data Distribution. Isn't it necessary that users can download cited data so they can reproduce experimental results? Just trying to make that explicit here.

rfcs/text/0000-data-citation-plan.md
theathorn: The implementation mechanism for User Stories 5 and 6 may differ. We may not require generic Data Portal login for the Data Ops team (e.g. by supplying specific internal tools to them) and the level of citation granularity may be restricted (e.g. to whole projects and/or the whole Data Distribution). User Story 6 allow for a wide audience of authorized users to create arbitrary cross-project citations.

rfcs/text/0000-data-citation-plan.md
jahilton: Yes, users must be able to download data from a Data Distribution, but it seems to be an unnecessary step when a simpler criterium would be "The Data Browser provides a means of downloading the data associated with a specific version of a project" (no matter where that version is mentioned)
I worry that adding unnecessary complexities are going to make implementation and acceptance more difficult to achieve. For instance, the current criterium relies on some functionality of Data Distribution, which has not been fully laid out yet.

rfcs/text/0000-data-citation-plan.md
jahilton: Similar with my comment below, I think assumptions about future functionality is introducing unnecessary complications (Data Distributions, DataOps logins, non-DCP user logins, collections). For instance, it has not been decided that all Data Distributions reference data at the 'project' level. 
Would a more straightforward (and thus, easier to spec out & implement) proposal be Phase 1: Stable non-versioned citable project URLs & Phase 2: Stable citable project _version_ URLs? It seems to me that a project version should be citable independent of any collection/distribution, so once Phase 2 is complete, those project versions could easily be grouped/listed in whatever packaging is required, whether by DCPers or non-DCPers.

hannes-ucsc (approved):

rfcs/text/0000-data-citation-plan.md
theathorn: Removed most dependencies on Data Distribution features.

rfcs/text/0000-data-citation-plan.md
theathorn: My intention was to state that for Phase 2 the Data Ops team decides when a project version becomes citable and creates the citable reference (DOI) for each version. Phase 3 then allows external users to create their own citations for any arbitrary set of data, I've attempted to remove any dependence on anything but the most  basic Data Distribution features.

lauraclarke: @theathorn I think that if we automatically make the headline url for a project citable then we should also automate citability of distinct project versions when we support access at that level regardless of why the version was updated, it will make the whole process much simpler and much more FAIR

@jahilton and the Data Ops team will, of course, end up needing to make decisions about how releases are updated and versioned but I agree with @jahilton that being too specific on mechanisms around that here is a bad idea

rfcs/text/0000-data-citation-plan.md
theathorn: @diekhans If your concerns have been addressed can you now approve this PR? Your status is still "requested changes".

diekhans (approved):
All good, really nice work.

rfcs/text/0000-data-citation-plan.md
theathorn: Removed all references to Data Distributions and replaced with "discrete project versions".

PR #106, opened 2019-08-14
[WIP] Fusillade Process Management for DCP

Bento007:


rfcs/text/Fusillade in the DCP.md
```diff
@@ -0,0 +1,105 @@
...
+
+1. Define the different types of users you need to support based on what actions and resource they will need permission for. These users types you define will be your roles. 
+
+1. Create a file **{role}.json** for each user type. Using the resource and actions previously define, write a policy in the AWS IAM policy format that express what each role can do. If it isn't explicitly stated, it is assumed the user with this role does not have permission. For the resource field, `*` can be used as a wild card in the same way it's used in AWS IAM policies.
+
+1. Run the script `??? -path "path to roles" -component "name of the component" -stage "current deployment stage" -secret "secretId of a service account in AWS" -service "path to service account credentials"` to add your roles to fusillade. The role that is added to fusillade will be named `{component}-{stage}-{role}`. The stage is not included for *prod* deployments.
```
DailyDreaming: `??? -path "path to roles" -component "name of the component" -stage "current deployment stage" -secret "secretId of a service account in AWS" -service "path to service account credentials"`

Guessing this should include a `-role "role_name"` option if it's going to be named `{component}-{stage}-{role}`.  I think it would be useful to warn users when a role with that name already exists, and maybe add a `--force` option to overwrite existing roles.  Also, kind of a nit, but I think I prefer `--` options rather than `-` just to be consistent with the `data-store`.

rfcs/text/Fusillade in the DCP.md
```diff
@@ -0,0 +1,105 @@
...
+        "the_roles_assigned_to_this_group"
+      ]
+}
+``` 
+
+1. In DCP infra run the script `??? -path "path to groups" -stage "current deployment stage" -secret "secretId of a 
+service account in AWS" -service "path to service account credentials"` to add your group to Fusillade. If any roles 
+specified in the group do not exists, an error will ne thrown and that group will not be create if it didn't exist, 
+or it will not be updated if it existed. Any member not in the group will be added to the group. The group added will
+ be named `{group}-{deployment}`. `{deployment}` is excluded from name when deployment is `prod`. For all deployments
```
DailyDreaming: `For all deployments the group named user_default does is not contain the deployment.` was unclear to me.

rfcs/text/Fusillade in the DCP.md
```diff
@@ -0,0 +1,105 @@
...
+        "a_list_of_user_emails@example.example",
+        "a_list_of_service_account_emails@example.example"
+      ],
+      "roles": [
+        "the_roles_assigned_to_this_group"
+      ]
+}
+``` 
+
+1. In DCP infra run the script `??? -path "path to groups" -stage "current deployment stage" -secret "secretId of a
```
DailyDreaming: Same as the above, I'm assuming there needs to be an argument for the unique group name and a check.

rfcs/text/Fusillade in the DCP.md
```diff
@@ -0,0 +1,105 @@
...
+1. As a component of the DCP, I would like to use service account to programmatically manage my service.  
+
+## Scientific "guardrails" [optional]
+
+*Describe recommended or mandated review from HCA Science governance to ensure that the RFC addresses the needs of the scientific community.*
+
+## Detailed Design
+
+### As a component of the DCP, how do I control access to my resources?
+1. Define what actions and resources you'd like to control. Actions should take the form `{component}:{Action}`, for example to get a bundle from the DSS the action could be `dss:GetBundle`. The action is case sensitive, and camel case is suggested. Resources are what actions affect. The same action can be applied to multiple different resources. For the resource use the AWS ARN format. For example...
```
DailyDreaming, line 160: Should there be a distinction between `dss:GetBundle` for AWS and GCP?

rfcs/text/Fusillade in the DCP.md
Bento007: The path would be to a folder which contains all of the roles, so they can be updated all at once. It's probably better to do it one at a time so you know what role failed. I'll add `--` to the parameters, `--role`, change `--path` to `--file`. Error returned if role exists, and `--force` will overwrite the role if the roles exists.

rfcs/text/Fusillade in the DCP.md
Bento007: My opinion is that this would be a field in the resource.

Bento007: Components have the freedom to determine the relationship between actions and resources. I will expand on the guide lines for defining resources, and add some examples.

rfcs/text/Fusillade in the DCP.md
Bento007: fixed

rfcs/text/Fusillade in the DCP.md
Bento007: fixed

rfcs/text/Fusillade in the DCP.md
Bento007: I added additional documentation for resources to help make this clear for components

PR #109, opened 2019-08-31
Proposing changes/updates to the RFC process.

maniarathi:
~~**September 24**: Last call for community review~~
~~**(EXTENDED) September 27**: Last call for community review~~
~~**October 4**: Last call for oversight review!~~
~~**October 17**: Up for approval at DCP PM meeting.~~
**October 17**: Approved at DCP PM meeting!

Summary of comments from community review: Addressed concerns around adding a Timeline section to the template (which has been removed), elucidating three main types of RFCs, and general editorial comments.

PR #113, opened 2019-09-16
Import pre-RFC-process-dated DCP Media Types

kislyuk:
This RFC imports the DCP Media Types design document (https://docs.google.com/document/d/1TqihrgXjct9aDmTJO52_gE2WlpFysB1OkG9C8exmWTw/edit#) that was decided upon prior to the RFC process.

**November 19th:** Last call for community review


sampierson (approved):

diekhans (approved):

kislyuk:
Given that this document pre-dates the formal DCP RFC process and is de facto approved and implemented, I plan to bypass the usual process and merge it in after letting it sit for a week. If you have objections or minor edits to propose, please let us know by the end of this week. Barring objections, I'll merge the doc in on Monday, September 23.

mweiden:
@kislyuk @diekhans please go through the standard RFC process for this. Not doing so sets a precedent that may lead to unexpected outcomes later: people might be lead to believe that they can bypass the process if they implement something first and _then_ publish the RFC.

If the RFC process is too slow, consider adding input to [RFC: Proposing changes/updates to the RFC process](https://github.com/HumanCellAtlas/dcp-community/pull/109).

sampierson:
Matt,

Did you understand the scenario here?
This is not about code that was written before there RFC process.
This is an RFC that was written before the current RFC process.
Are you suggesting that every time we tweak the RFC process, we need to re-review all RFCs?

-Sam.

sampierson:
Ping @mweiden 
Regarding my comment above.
This RFC was reviewed by the Arch team in October 2017, using the Google Docs review mechanism.  Review was started here https://humancellatlas.slack.com/archives/C6SGU0R3J/p1508792188000502
Do you still want it to be reviewed again?

mweiden:
Discussed during the 2019-10-11 Arch meeting. We reached consensus on importing all old RFCs from google drive and labeling them "Imported."

diekhans:
I believe it is important to put all labels at the start of the document as well as github labels so it is easy to see when reading.

diekhans (changes requested):
If we are going to import these into the current process, lets really import them by converting to markdown and committing.  This way they can be used and maintained as with all other RFCs.
It doesn't really need to follow the template exactly, just be under source control and follow the naming convention.

This can be semi-automated by saving the page as HTML and doing

  `pandoc DCPMediaTypes.html -t markdown_github > DCPMediaTypes.md`

Although the markdown is a bit ugly and needs some editing.  Probably other tools too.

diekhans (changes requested):
Above duplicated by github error 500.

diekhans (changes requested):
Above duplicated by github error 500.

diekhans (changes requested):
Above duplicated by github error 500.

diekhans (changes requested):
Above duplicated by github error 500.

maniarathi:
Updated to import DCP Media Types again.

rfcs/imported/0000-dcp-media-types.md
```diff
@@ -0,0 +1,149 @@
+### DCP PR:
+
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC number and this pull request in the format:*
+`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+
+# DCP Media Types
+
+DISCLAIMER: Please note that this RFC has been imported its original Google Doc and went through a design review process prior to the implementation of the current [RFC Process](https://github.com/HumanCellAtlas/dcp-community/blob/master/rfcs/text/0001-rfc-process.md). The original document is [here](https://docs.google.com/document/d/1TqihrgXjct9aDmTJO52_gE2WlpFysB1OkG9C8exmWTw/edit#heading=h.87ix45a71erf). Please be aware that the contents below may potentially be out-of-date as the last-modified date of the original Google Document is October 20th, 2017.
```
sampierson: Your "out of date" comment seems unnecessary.  They same could be said of all RFCs the day after they are written.
And AFAICT this information is all still current at this time.

rfcs/imported/0000-dcp-media-types.md
```diff
@@ -0,0 +1,149 @@
...
+
+[Andrey Kislyuk](mailto:akislyuk@chanzuckerberg.com)
+
+## Shepherd
+
+[Arathi Mani](mailto:arathi.mani@chanzuckerberg.com)
+
+## Motivation
+
+The DCP should not invent new media types; we should append `dcp-type=metadata` or `dcp-type=data` to our Content-Types using the media type "parameter" syntax, e.g. `Content-Type: application/json; dcp-type=”metadata/sample”`. Instead the DCP should strive to use "in-band" media type communication mechanisms where possible.
```
sampierson: This paragraph seems a little mangled.  I think what you are trying to say is "The DCP should not xxx  . Instead it should do yyy".  Right now the Instead clause come after the "should do" clause.

sampierson (changes requested):
This paragraph seems a little mangled.

rfcs/imported/0000-dcp-media-types.md
diekhans: Agreed; we should be checking for accuracy on import.  I think the important thing to note  is not that is was in google docs, but it was approved by an older process.

It would be good to update the google doc to point to the RFC.

diekhans:
note: I will be importing the domain RFC ~ Nov 13th

rfcs/imported/0000-dcp-media-types.md
maniarathi: @sampierson Deleted the last sentence. Thanks.

@diekhans I have an earlier sentence in that paragraph "and was approved by a design review process prior to the implementation of the current RFC Process."

@diekhans @sampierson I will let the original author decide whether to point the Google Doc to the RFC :)

rfcs/imported/0000-dcp-media-types.md
maniarathi: Done.

diekhans:
I don't see the value in putting these into a special directory "imported".  This just creates a special case because of being reviewed with a different process.   I think the idea is for these to be first-class RFCs just like all the others.  We don't really want to have a new directory every time we change the RFC process.

sampierson:
> I don't see the value in putting these into a special directory "imported". This just creates a special case because of being reviewed with a different process. I think the idea is for these to be first-class RFCs just like all the others. We don't really want to have a new directory every time we change the RFC process.

I agree with you Mark.  Arathi however seems to be on a mission to keep this RFC from every being a peer with the post-documented-community-process RFCs.  This is kind of insulting, but I am so sick of this topic that in the spirit of compromise I decided to let her do it.  It makes developers lives harder, but not by much, and we have more important things to be thinking about.

PR #111, opened 2019-09-03
Versioned Experimental Data Structure Conceptual Model

diekhans:
This informational RFC described the conceptual model of the HCA DCP experimental data and associated metadata in light of multiple versions of data.

Status: Paused waiting on data model working group
Last call for community review:

PR #104, opened 2019-08-12
Data Operations charter

jahilton:
~~**August 26**: Last call for community review~~
~~Revisions in progress~~
Summary of revisions:
- Specified high-level objectives
- Clarified and added definitions to align (or distinguish) with existing DCP terminology
- Clarified Data Operations role relative to other DCP components

~~**September 20**: Last call for oversight review~~
~~Final revisions in progress~~
Summary of revisions:
- Aligned with (more prevalent) existing DCP terminology (submission)
- Clarified Data Operations role in connecting community feedback to the DCP components

PR #115, opened 2019-09-17
RFC project completion stage

jahilton:
~**October 1**: Last call for community review~
Summary of revisions:
- Broke down functionality specifications into phases
- Defined _version-complete_ and how a project becomes _version-complete_
- Added Impact of past/current errors

~**October 9**: Last call for oversight review~
Summary of revisions:
- Laid out acceptance criteria, unresolved questions, and drawbacks/limitations


diekhans:
This RFC should be written in the context of the mechanism in [Processing Datasets that Span Multiple Data Collection Runs](https://github.com/HumanCellAtlas/dcp-community/pull/88).  If new requirements are created by this RFC, updating the #88 will be appropriate.

rfcs/text/0000-project-completion-stage.md
```diff
@@ -0,0 +1,70 @@
...
+
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC number and this pull request in the format:*
+
+`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+
+# Project completion stage RFC
+
+## Summary
+
+This RFC proposes a process of releasing a contribution, or project, only after it has been completed by the DCP in order to maximize the value to the data consumer through reducing user confusion and minimizing errors in the final product.
```
theathorn: This is fundamentally different from how the Data Store and Data Browser currently operate (i.e. processing bundle by bundle).  Does the DCP support a “live” vs “hybrid” vs “curated data” presentation of data? This seems like a fundamental question that must be answered first.

jahilton: I would love clear definition of what people think live/curated/hybrid models are, and an explanation of how this RFC would impact any of that.
The DCP is not currently doing any curation, nor is any proposed in this RFC.

theathorn: How about "batch" or "embargoed" model instead of "curated"? Looking for a one word description of what this RFC is proposing. - i.e. "display only completed data".

diekhans: "embargo" has a very specific meaning and batch sounds rather techie, like the punched cards @theathorn use to carry around.

From a user perspective,  it is about the completion of an experiment.  The fact that we chop it up into a batch of bundles is something they shouldn't care about.  An experiment comes in and the results should come out as a whole not tweeted out in bundles.

I was using the term "complete submission"  however that caused confusion.  I was continually reminded that nothing is complete, as it can be updated, so I started using "version-complete", which was clear.   Then submission caused confusion since submission means a very specific thing in ingest and is not how consumers really think about obtaining data.  I believe people want the results of an experiment, although we don't yet have an experiment concept in the DCP, there is a discussion about adding it.

I see the concept as "version-complete experiment",  but probably needs simpler name, however, I am at a loss for that name.

jahilton: Many people associate "embargoed" with databases that hold data private for months while the contributor publishes.
I'm also going to get nit-picky on "live". We're not even sniffing the data until it's months or years old.
No matter what you call it, I'm not understanding how this changes the DCP model. Currently, the wranglers and the submitters go back-and-forth for weeks correcting, refining, validating submissions - all completely hidden from the public view. This RFC is just proposing that we extend that period to include the data lifecycle within the whole of the DCP, allowing us to correct, refine, validate _all_ DCP products.

diekhans: Currently, everything goes "DCP-live' bundle by bundle, sans maybe timing issues on Azul updates.

So we can have 1/2 of the primary data in the browser and part of the with analysis.

I don't really know what happens with stating before submitting to the DSS

So I don't think this is changing anything fundamental, more than formalizing it.  We currently don't even have the terminology to talk about such things.

rfcs/text/0000-project-completion-stage.md
```diff
@@ -0,0 +1,74 @@
...
+- Project ingested at the wrong schema version ([#2421](https://github.com/HumanCellAtlas/data-store/issues/2421))
+- Project Metadata tsv download threw an error
+- Contributor name misspelled ("Tom Mithcell")
+
+As a contributor whose name is associated with the project, I want to be allowed time to review the full details of my contribution as for accuracy before it becomes publicly available. I want this review to be through the vantage of a potential user. (possibly 2nd phase)
+
+As a contributor, I am curious where my submission is in the DCP’s data lifecycle after I have provided the metadata spreadsheet all the way until it is released. (possibly 2nd phase)
+
+
+## Detailed Design
```
NoopDog: Jason, thanks for preparing the RFC. I would like to highlight a few key use cases and propose some additional project lifecycle states and related project views to support them.

## Use Cases
1. As a project contributor, I would like to sanity/science check  my data before making it available to all users.

1. As a project contributor, I would like to use the DCP to analyze my data and I want to be able to review the completed analysis in the system without having to wait for it to be included in an HCA Release.

1. As the DCP I would like as many eyeballs on the data as possible, as early as possible in the project lifecycle (e.g. before a project is part of an official HCA Release), to support the early detection of errors and insights.

1. As a data consumer, I would like early access to version-complete projects that are not yet part of an official HCA Release.

## Project Lifecycle Definitions

- **Version-incomplete** - Project is being processed, best effort real time.
- **Version-complete** - All expected project artifacts (primary, secondary, matrix) are complete and pass a TBD basic QC check.
- **Released** - The project is version-complete and belongs to a "Capital R" Release.

## Project Views
### Preview / Staging View
- DCP Admins, DCP Wranglers can see the latest version of all projects including incomplete projects.
- Project contributors can see their own version-incomplete projects.
- Project contributors can see their own embargoed projects.
- Processing is best effort real time, bundle by bundle as data is added. 
- Supports QA/QC efforts.

### Leading Edge View
- Shows  the latest version of version-complete and released projects only.
- Allows for early access to unreleased released, version-complete projects.

### Release View
- Shows released projects only.
- Visible by all.

## Project Versions/ Big R Releasees
Implied in the above is that projects have versions and that Big R Releases are composed of versioned projects. A first pass of how this might work is as follows:

### Project Versions
A project would have an initial version on its first submission of data.

The following events would bump the project version:
- Adding a new data submission.
- Adding a new pipeline or analysis tool that created a new project product.
- Adding a new version of a pipeline or analysis tool that re-ran the pipeline updating a project product.
- Removing data (as in some TBD curation process) 

For example, a projects history as displayed at some point on the project detail page might look like:

_Project X_
Version 3 - Removed bad run of mouse data.
Version 2 - Mouse SS2 pipeline bug fix
Version 1 - Mouse SS2 pipeline Released
Version 0 - Initial upload of human and mouse SS2 data

### Big R Releases
A Release can contain many projects but only a single version of each project.
A project version can be in multiple Releases.

An example of 3 Releases over 2 projects Project X and Project Y might look like:

HCA Release 3 - Project X V3, Project Y V2
HCA Release 2 - Project X V2, Project Y V1
HCA Release 1 - Project X V1,  Project Y V1

**Notes:**
The "Preview/Staging View" most closely matches what we have now except that currently, all projects in any state are visible to anyone.

In terms of incremental adoption we could add the "Leading Edge" view next and then add authentication/authorization to restrict the "Staging View"

There may be better names for the states please suggest. 

I added initial detail about project versioning and releases to give some context about where we might be heading.

jahilton: Thanks @NoopDog!

> As a project contributor, I would like to QC my data before making it available to all users.

This is a touchy one that starts to wade into embargoing, which we definitely want to avoid. What happens if the contributor is not please with the QC? I'm pretty sure there's a "no take backs" clause in the terms & conditions.
Or does this refer to the 'contributor sanity check' in my 7th User Story (which is about data _accuracy_, not data _quality_)?

I think there's some terminology mis-alignment. To 'release' means to make it accessible to data consumers while what you call a 'release' is really a 'Snapshot' ([DataOps charter](https://github.com/HumanCellAtlas/dcp-community/pull/104) for reference). This may clarify why in the places you mentioned "early access to unreleased data", I would consider that to be "access to data".

But your concept of a Preview/Staging area seems very much aligned with the proposed 'signed-in' view (and sounds better!), same with Leading Edge == what we currently have & the proposed 'signed-out' view.
I think your concept of 'Release View' will be incredibly relevant once we start with Snapshots. I am hesitant to work that in with this RFC in order to keep this limited to 'improving what we currently have', but I do recognize that come implementation time, it may behoove us to have each 'view' spec'd in order to maximize engineering efficiency.

rfcs/text/0000-project-completion-stage.md
```diff
@@ -0,0 +1,74 @@
...
+## Shepherd
+***Leave this blank.** This role is assigned by DCP PM to guide the **Author(s)** through the RFC process.*
+
+*Recommended format for Shepherds:*
+
+ `[Name](mailto:username@example.com)`
+
+## Definitions
+
+A contribution, or project, is considered to be **version-complete** when the contributor confirms (by way of the wrangler) that they have provided all of the metadata and raw data that they plan to submit at the current time, the wrangler submits those to ingest, and the DCP has delivered every possible product, given current DCP capabilities.
```
tburdett: I like the spirit of this. But:

> contributor confirms (by way of the wrangler) that they have provided all of the metadata and raw data that they plan to submit at the current time

...seems to imply addition of extra process during ingestion, when in fact this is simply what the ingest "submit" button is for. This can be made much simpler:

```suggestion
A contribution, or project, is considered to be **version-complete** when the DCP has delivered every possible product, given current DCP capabilities, for a given submission.
```

tburdett: I find the "contribution or project" nomenclature very fuzzy. What's the real concept (or maybe aggregated set of concepts) that you care about here? All the data and metadata from a single project?

Could you name that concept for the purposes of this document and then create a definition? It might also be worth talking with @gabsie as she is trying to figure out how to write a "dataset RFC" and I think you might be thinking of the same concept.

PS. It would make me happy if you don't use "project" for this concept, because that tends to overload the usage in the metadata model.

rfcs/text/0000-project-completion-stage.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Motivation
+
+**Display only completed data:**
+Any project that contains raw data with a corresponding analysis pipeline should be considered in progress until all downstream analysis is version-complete, validated, and available. Currently, a contributed project is made accessible to the public as soon as raw data & metadata are submitted. The trickling in of data can lead to data consumer confusion as to if/when more data will be added to the project. The proposed solution allows the DCP to complete the version of the project with DCP outputs before the project becomes publicly accessible.
+
+**More thorough data validation:**
+The downstream services of the DCP have the potential to uncover metadata and data errors. Viewing the version-complete project as the user will view it can also identify errors. A solution to this is a setting on projects allowing signed-in DCP members to access the data in order to perform a final ‘sanity check’.
```
tburdett: SO MUCH THIS!

For me, the greater value here is around coherent QA and QC for entire projects (and yes, I am using these terms even though I know they're ambiguous). I think this is more important than detection of errors in the metadata and data. Metadata and data errors will always be present - we'll never build a perfect data/metadata validation system without bugs - and I'd further suggest the cost/benefit ratio of having a human review these might not be there.

However, I **totally** agree that having the possibility to automate checks that projects meet desired QC/QA standards before making them public, and having the chance to manually spot anything obvious allows us to be much more responsive in detecting and repairing errors in the data or bugs that creep in during ingest/storage/analysis/presentation processes.

rfcs/text/0000-project-completion-stage.md
```diff
@@ -0,0 +1,74 @@
...
+
+As a member of the DCP, I cannot do a comprehensive final check on data on the DataBrowser or DataStore unless I see it as the data consumer will see it.
+
+As a member of the DCP, I do not want inaccurate metadata or products of incorrect data analysis being shown to the public. 
+Examples of inaccuracies that were/are publicly accessible:
+- Optimus results on Smart-seq2 data ([#2314](https://github.com/HumanCellAtlas/data-store/issues/2314))
+- ‘Homo sapiens’ vs ‘homo sapiens’ ([ZD ticket 158](https://humancellatlas.zendesk.com/agent/tickets/158))
+- Project ingested at the wrong schema version ([#2421](https://github.com/HumanCellAtlas/data-store/issues/2421))
+- Project Metadata tsv download threw an error
+- Contributor name misspelled ("Tom Mithcell")
```
tburdett: See my comment above for thoughts on this. Some of these examples are trivial (e.g. the typo in contributor name) and have little impact on the value provided by the DCP, whereas others are significant issues related to data correctness (QA/QC) that massively impact value to users.

Generally, I'd rather we focus on adding the ability to review for very significant problems in the data before going "live" over catching trivial types of errors, and further I'd prefer us to responsively fix trivial errors that creep into production rather than introduce processes to try to catch them all up front. As such, I would refocus these comments to consider impact of errors, rather than simply listing past mistakes.

rfcs/text/0000-project-completion-stage.md
tburdett: I like the "version-complete" concept. By ingest design, submissions **are** complete once they are submitted, though. To add more data to a project requires a second submission. See my suggestion, below, for more on this.

And, I would do more than nit-pick on "live". The idea that anything is "live", simply because we have locked together and automated the processing and release of bundles, is a fallacy. We expend considerable effort chasing this level of automation when in fact data and metadata is bottlenecked for months before even first contact with the DCP. Even _after_ first contact, it is still weeks whilst researchers in labs prepare spreadsheets and data uploads (as @jahilton nicely describes). Once prepared, data arrives in coarse-grained submissions. Arbitrarily breaking apart these coarse-grained but useful units and streaming the pieces without providing any signal of how many to expect is worse than useless.

tburdett (changes requested):
I've left a small number of comments for which I am requesting changes - I think there's places where you could considerably strengthen the clarity of this RFC. But they should all be small editorial fixes; the general thrust and concepts in this RFC are fantastic.

rfcs/text/0000-project-completion-stage.md
ambrosejcarr: I think this is the key point: 

> So I don't think this is changing anything fundamental, more than formalizing it

I also think it's worth recording the historic context: When the DCP was conceived, it **was** thought that users would bring data directly from their sequencers to the DCP, and many decisions were made with aims to support such a workflow. These decisions included separating SS2 experiments into independent bundles and supporting "live" workflows. As Tony notes above, we've since observed that this assumption was wrong, and this RFC formalizes those observations. In so doing, it balances the advantages of rapid processing against data quality assurance.

claymfischer: @ambrosejcarr put it very succinctly. +1 to version-complete, here.

rfcs/text/0000-project-completion-stage.md
diekhans: I find project confusing because of the existing concept of BioProject.  Essentially related to a  grant or research group.

I find "dataset' by itself confusing, as it is very general and lot of people think they know what it means (GA4GH 1.0 horror story available on request).

I believe "experiement' or "experiment dataset" is much clearer and maps to how people think about science.

We also need to think how this terminology maps to primary and secondary analysis.

Maybe the whole mess is an experiment and with primary results are an experiment dataset, etc.

We really need to turn @justincc terminology table into an RFC.

lauraclarke: I want to second Tony's proposal, I think it better reflects current work and also means the text won't need to be as we move more of the submission process to be contributor-led rather than wrangler-led

lauraclarke: Depending on which archive/resource you look at, project, experiment, study and dataset are used to represent overlapping concepts. We picked project because the concept of the logical overarching set of data needed a label. If it is useful to change it we can.

I personally dislike experiment as in my head that represents the individual unit of work and output rather than the collection

That said if we get too concerned with avoiding terms that are understood in different contexts to mean different things we might as well pick a completely arbitrary word like banana and just give up trying to pick a word which is close in meaning already

The most important issue here is that we define what we mean when we use these words rather than necessarily change words (though if there is a consensus for a different term for that top-level grouping to project I won't stand in its way)

rfcs/text/0000-project-completion-stage.md
lauraclarke: This, I am very concerned that this will unduly slow down making data available to the public if we put too many humans in the way. Seeking to automate as much checking and accepting that we will never catch 100% of errors but we can get better overtime seems like the best compromise approach

rfcs/text/0000-project-completion-stage.md
```diff
@@ -0,0 +1,74 @@
...
+
+**More thorough data validation:**
+The downstream services of the DCP have the potential to uncover metadata and data errors. Viewing the version-complete project as the user will view it can also identify errors. A solution to this is a setting on projects allowing signed-in DCP members to access the data in order to perform a final ‘sanity check’.
+
+**Support data contributors (possibly 2nd phase):**
+Allowing data contributors to view the final product in their collaboration with the DCP before it is released adds yet another validation step (after all, who knows their data better?) and is a valuable step in good faith data sharing. The inclusion of a data lifecycle tracker on the project page keeps the contributor up-to-date on their submission and ensures a transparent process.
+
```
lauraclarke: While I like the idea of giving a contributor a chance to review we must also be careful we don't set an expectation that a contributor can block the release of data if they disagree with the results of our pipeline or just want to delay other people getting hold of the data. Such a release will need to have a timebox associated with it and passive acceptance being assumed at the end of that timebox

rfcs/text/0000-project-completion-stage.md
```diff
@@ -0,0 +1,74 @@
...
+As a member of the DCP, I do not want inaccurate metadata or products of incorrect data analysis being shown to the public. 
+Examples of inaccuracies that were/are publicly accessible:
+- Optimus results on Smart-seq2 data ([#2314](https://github.com/HumanCellAtlas/data-store/issues/2314))
+- ‘Homo sapiens’ vs ‘homo sapiens’ ([ZD ticket 158](https://humancellatlas.zendesk.com/agent/tickets/158))
+- Project ingested at the wrong schema version ([#2421](https://github.com/HumanCellAtlas/data-store/issues/2421))
+- Project Metadata tsv download threw an error
+- Contributor name misspelled ("Tom Mithcell")
+
+As a contributor whose name is associated with the project, I want to be allowed time to review the full details of my contribution as for accuracy before it becomes publicly available. I want this review to be through the vantage of a potential user. (possibly 2nd phase)
```
lauraclarke: Like with contributor review of analysis results we need to make sure if such a mechanism is introduced that it doesn't establish the expectation that all named scientists in a project get to actively okay a project submission. The current system trusts the person directly interacting with the wrangler to be able to give permission for all named scientists and I expect that will continue when that contributor is directly interacting with the system.

Within ingest we have a long term vision which we haven't yet realised which would allow different participants in the project lifecycle being able to enter the data or metadata they are responsible for (tissue acquisition potentially being different from assay details) but even if that is possible it is still most likely that for most submissions only one person will interact with us and that should be sufficient. Hopefully, our authentication and authorization system will ultimately allow multiple people to log in and view but it shouldn't be required

rfcs/text/0000-project-completion-stage.md
```diff
@@ -0,0 +1,74 @@
...
+
+- A mechanism for DCP members to sign-in to the DataBrowser and DataStore.
+- A mechanism for a project to be visible only for a user that is signed-in.
+- Ensure Secondary Analysis and Matrix Service are able to access the data hidden from signed-out users and to contribute their products back to the project
+- A mechanism for Data Operations to switch a project from not visible for signed-out users to visible for all users (data release).
+- A ‘checklist’ and validation procedure for Data Operations to perform on each project within a week of being deemed version-complete.
+- A mechanism for a data contributor to sign-in to the DataBrowser and DataStore, and be given access their own project before it is visible for all users. (possibly 2nd phase)
+- A dashboard on the project page for signed-in users to track where a given project is in the data lifecycle. (possibly 2nd phase)
+
+*The portions of the proposal addressing contributor access to only their projects have been noted as “possibly 2nd phase”. In anticipation of a logical breakpoint in technical effort required, the implementation of DCP-only access should not be delayed by these additional requirements.*
```
lauraclarke: This is a long least of requirements. I think it would be helpful for the engineering plan to define which business requirements must be serviced by the MVP solutions for this process and which can be delayed in more specific manner than this. I think the phased requirement list in the citation RFC does this well.

rfcs/text/0000-project-completion-stage.md
NoopDog: >That said if we get too concerned with avoiding terms that are understood in different contexts to mean different things we might as well pick a completely arbitrary word like banana and just give up trying to pick a word which is close in meaning already.

Exact.

One idea is to namespace the word, for example "HCA Submission". We might also just accept that there will be some training about what a given word means in the HCA context.

lauraclarke: Adding HCA as a prefix to help understanding and provide context is a good idea.

That said, we need to be careful not to overload the word submission.

A submission is an event where a user interacts with the ingest system. A project might be made up of multiple submission events. So I think we need to pick a different word to that

rfcs/text/0000-project-completion-stage.md
justincc: > The idea that anything is "live", simply because we have locked together and automated the processing and release of bundles, is a fallacy. We expend considerable effort chasing this level of automation...

I would go beyond "considerable" effort. Our entire technical architecture, and the resulting complexity, has been shaped by the requirement to stream data out as quickly as possible.

rfcs/text/0000-project-completion-stage.md
diekhans: > The most important issue here is that we define what we mean when we use these words rather than necessarily change words

The rigorous definition of terminology is the key, otherwise, people make assumptions based on their experience.

jahilton: Thanks for the suggestion...though I would like to reserve the right to update the "submission" term after I wade through the discussion below

rfcs/text/0000-project-completion-stage.md
jahilton: Fully agree on all of this. Our goal is to improve, not catch _everything_.
We set the review period at 1 week max knowing that it will likely be fully manual to begin. But the goal will be to automate the DataOps checklist/validation procedure as soon as possible.

rfcs/text/0000-project-completion-stage.md
jahilton: Great suggestion. I have added the impact of these past mistakes in 28fadb9.
I will disagree that a contributor name typo is trivial. In terms of FAIR data and data accuracy, you are correct, it has no impact. But it's a really bad look for the DCP, especially at these early stages as we are seeking credibility in the community.

rfcs/text/0000-project-completion-stage.md
jahilton: Agreed, I have clarified that the contributor review is to be confined to the same week timebox as the internal DCP review. b175eb3

rfcs/text/0000-project-completion-stage.md
jahilton: Excellent point. I have added that clarification in b175eb3.

rfcs/text/0000-project-completion-stage.md
jahilton: I have more clearly distinguished the functional specifications between Phase 1 & 2 in ca09319

rfcs/text/0000-project-completion-stage.md
jahilton: There's some inconsistency on whether a 'contribution' or 'submission' is provided to the DCP by data contributors/submitters (convo on the DataOps Charter). I have made this RFC internally consistent (submission/submitter for now and will update if the Charter convo goes the other way).
To me, _project_ has a very clear definition at the DCP & HCA. It is the primary organization unit that comes in to the DCP and that goes out to the user (click 'explore' on the portal and 'projects' is in the url, it's a tab label, click a _project_ title and you get to explore _project_, or select _project, or download the _project_ metadata).
I do agree that adding 'HCA' is a fabulous (and very necessary) idea - hopefully that distinguishes it from how other ~projects~ groups define 'project'.
UPDATED - Hopefully the included definitions and more consistent use convey my meaning better now, but let me know if there's still uncertainty - thanks to all for this!

rfcs/text/0000-project-completion-stage.md
jahilton: Even more Phase break down in eef5941

rfcs/text/0000-project-completion-stage.md
jahilton: Reading through the updates to this comment...
I think there's some overload in to this one proposal.
There's some discussion on versioning - There's a great [Informational RFC](https://github.com/HumanCellAtlas/dcp-community/pull/111) on how that would be modeled
There's also discussion about HCA Data Snapshots that the DataOps team will get to, but is independent of the project completion proposal.

rfcs/text/0000-project-completion-stage.md
justincc: Does a project include all analysis generated by the DCP?

jahilton: @justincc Yes. I updated the definitions in c118d05.
I'll resolve this conversation, but please comment on the revisions if there's still questions/comments.

rfcs/text/0000-project-completion-stage.md
```diff
@@ -0,0 +1,94 @@
...
+As an HCA submitter whose name is associated with the HCA project, I want to be allowed time to review the full details of my HCA submission as for accuracy before it becomes publicly available. I want this review to be through the vantage of a potential user. (possibly 2nd phase)
+
+As an HCA submitter, I am curious where my HCA submission is in the DCP’s data lifecycle after I have provided the metadata spreadsheet all the way until it is released. (possibly 2nd phase)
+
+
+## Functional Specifications
+**Phase 1**
+*MVP requirements*
+- A mechanism for DCP members to sign-in to the DataBrowser and DataStore.
+- A mechanism for an HCA project to be visible only for a user that is signed-in.
```
TimothyTickle: The portal is not the only access point in the system. I believe this would require authN/Z to all API endpoints and for those endpoints to censor data based on the state of the HCA project.

rfcs/text/0000-project-completion-stage.md
TimothyTickle: I would not use "embargo" for these functionalities, this is still a live system just controlled in the release and understandable in its state. We may need embargo at some point but we do not have use cases here to motivate that level of delay.

rfcs/text/0000-project-completion-stage.md
TimothyTickle: I am all for automating the QA for data access, we need it. This is clear in recent months. curation by QC is something that makes sense for distributions but there should always be ways of accessing data that passed QA but not QC. That data is valuable to some use cases and may be recovered by clever people.

TimothyTickle (approved):
There has been a lot of great activity here that has moved this proposal from something I would have been concerned about something we absolutely need. I am excited to see the DCP work this way.

rfcs/text/0000-project-completion-stage.md
jahilton: This is great discussion and it seems to have reached a plateau. I will resolve this thread. Please renew comments if there are outstanding issues that I can address in the RFC

rfcs/text/0000-project-completion-stage.md
jahilton: Resolving this thread, please renew comments if I can address things further in the RFC.

rfcs/text/0000-project-completion-stage.md
jahilton: Resolving this thread, please renew comments if I can address things further in the RFC.

rfcs/text/0000-project-completion-stage.md
jahilton: Resolving this thread, please renew comments if I can address things further in the RFC.

rfcs/text/0000-project-completion-stage.md
jahilton: Resolving this thread, please renew comments if I can address things further in the RFC.

rfcs/text/0000-project-completion-stage.md
jahilton: I think we're all in agreement here. Our goal is to get proper QA checks in place first, and then we'll iteratively work at automating more and more to be more efficient.
Resolving this thread, please renew comments if I can address things further in the RFC.

rfcs/text/0000-project-completion-stage.md
jahilton: Resolving this thread, please renew comments if there's anything specific to the details in the RFC that I can address further. (I know it's difficult to avoid too much implementation details at this point - but very soon!)

rfcs/text/0000-project-completion-stage.md
jahilton: I think we're aligned in our ideas here. A likely next step if/when the RFC is accepted will be a focused call (or F2F session) with all the right people present and figure out implementation.

rfcs/text/0000-project-completion-stage.md
```diff
@@ -0,0 +1,94 @@
...
+  - Impact: Inconsistent access experience across HCA projects (a user specifying one value will get results, but not all desired results).
+- HCA Project ingested at the wrong schema version ([#2421](https://github.com/HumanCellAtlas/data-store/issues/2421))
+  - Impact: Inconsistent access experience across HCA projects. Resulted in breaking data permanency as the data were removed and resubmitted with new uuids and HCA project page url with no record of provenance.
+- HCA Project page Metadata tsv download threw an error
+  - Impact: A non-functioning user-accessible link. Inconsistent access experience across HCA projects if the metadata tsv is not available for all HCA projects.
+- HCA submitter name misspelled ("Tom Mithcell")
+  - Impact: A (rather personal) HCA submitter-facing error will put DCP credibility at risk.
+
+As an HCA submitter whose name is associated with the HCA project, I want to be allowed time to review the full details of my HCA submission as for accuracy before it becomes publicly available. I want this review to be through the vantage of a potential user. (possibly 2nd phase)
```
lauraclarke: It would be good to see the possiblys turn into real goals, either there is an intent for a phase 2 or there isn't

lauraclarke:
The absence of all of the later sections in the RFC template in this document concerns me?

https://github.com/HumanCellAtlas/dcp-community/blob/master/rfcs/rfc-template.md

`Acceptance Criteria [optional]`
`Acceptance criteria are the conditions that a RFC must satisfy to be accepted by users or other` `stakeholders.`

`Unresolved Questions`
`What aspects of the design do you expect to clarify further through the RFC review process?`
`What aspects of the design do you expect to clarify later during iterative development of this RFC?`
`Drawbacks and Limitations [optional]`
`Why should this RFC not be implemented?`

`Prior Art [optional]`
`Share references to prior art to deepen community understanding of the RFC, such as learnings,` `adaptations from earlier designs, or community standards.`

`Alternatives [optional]`
`Highlight other possible approaches to delivering the value proposed in this RFC. What other` `designs were explored? What were their advantages? What was the rationale for rejecting` `alternatives?`


Are there really no acceptance criteria or unresolved questions? is this an approach without drawbacks or alternatives?

rfcs/text/0000-project-completion-stage.md
jahilton: Excellent call. I floated it initially ready to drop it if there was significant objection. However, I haven't heard much objection, and I believe it does add value. Solidified this in 2382eb877fc250f93a665d05d688e30363755db9

jahilton:
@lauraclarke added more details in 66e00e0d60f1aff230d2c8790994fe40d2487ca3

tburdett (approved):
You've addressed all my concerns with subsequent changes, thanks Jason

lauraclarke (approved):
Thanks, looking forward to seeing this moving foward

PR #117, opened 2019-09-17
Deletion RFC updates

natanlao:
General changes
* Correct API paths for consistency with current API (e.g. `s/file/files`)

## Adding Deletion APIs
* Replace HTTP 201 with HTTP 202 to indicate physical deletion. Because physical deletions are processed in batches (and not performed at request time), HTTP 202 Accepted is more appropriate than HTTP 201 Created.
* Replace HTTP 404 with HTTP 410 in the event that a user attempts to delete a file that has already been deleted or queued for deletion. HTTP 404 is an inappropriate response because, in this case, the server is aware of the fate of the file/bundle in the request. (Since tombstones are not deleted with a physical deletion, we can detect this case reliably.)
* Return HTTP 404 in that case that a user attempts to delete a file or bundle that does not exist.

## Restore APIs
* Add case to return HTTP 409 Conflict if a user attempts to delete a file or bundle that currently exists.
* Replace HTTP 404 with HTTP 410 in the event that a user attempts to restore a file that has been permanently deleted. (As tombstones are not deleted with a physical deletion, we can detect this case reliably.)
* Return HTTP 404 in the case that a user attempts to restore a file or bundle that does not exist.

### Last Call Sept.  29th, 2019

### DIFF
--- a/rfcs/text/0004-dss-deletion-process.md
+++ b/rfcs/text/0004-dss-deletion-process.md
@@ -95,11 +95,12 @@ deletion process. The request follows this swagger document:
 
 |Response Code| Description|
 |--------------|------------|
-|200| Deletion pending. A confirmation code is returned to confirm the deletion in the second request. A list of affect files and bundles is also returned |
-|201| Deletion Confirmed with confirmation code. |
+|200| Deletion pending. A confirmation code is returned to confirm the deletion in the second request. A list of affected files is also returned. |
+|202| Confirmation code accepted, physical deletion pending. |
 |403| Unauthorized user it attempting this action.|
-|404| The files has already been deleted.|
+|404| The file does not exist. |
 |409| The confimation code used is invalid. Either the confirmation code was incorrect, or an effected bundle has changed state since the original request.|
+|410| The file has already been deleted.
 
 A user must have explicit permission to perform the file deletion operation.
 
--- a/rfcs/text/0004-dss-deletion-process.md
+++ b/rfcs/text/0004-dss-deletion-process.md
@@ -258,11 +258,12 @@ procedures takes place:
 
 3) The users makes a second request that included the confirmation code. This starts the deletion process.
 
-4) After the deletion request has been confirmed the DSS places **tombstones** in the
-   underlying storage system(S3 or GCP bucket). For a **logical deletion** request, tombstones are placed for all bundles
-   mentioned in the requests response. For **physical deletion** requests, tombstones are placed for all files and
-   bundles mentioned in the request response. The tombstone's content is
-   as follows:
+4) After the deletion request has been confirmed, the DSS places **tombstones** in the
+   underlying storage system (S3 or GCP bucket). For a **logical deletion** request,
+   tombstones are placed for all bundles mentioned in the response to the deletion request.
+   For **physical deletion** requests, tombstones are placed for all files and bundles
+   mentioned in the response to the deletion request. The tombstone's content is as follows:
+
    ```JSON
    { "reason": "<reason>",
      "details": "<additional info>",
--- a/rfcs/text/0004-dss-deletion-process.md
+++ b/rfcs/text/0004-dss-deletion-process.md
@@ -11,7 +11,7 @@ Establish a secure and controlled process for deleting data from the Data Store.
 ## Author(s)
 
  [Trent Smith](mailto:trent.smith@chanzuckerberg.com)
- 
+
  [Hannes Schmidt](mailto:hannes@ucsc.edu)
 
 ## Shepherd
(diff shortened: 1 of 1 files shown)


rfcs/text/0004-dss-deletion-process.md
```diff
@@ -64,77 +66,84 @@ Possible reasons for deletion of (meta)data are
...
-deletion process. The request follows this swagger document:
+While a user can only opt to **physically delete** a file with the file deletion
+API, a file can be considered **logically deleted** in the grace period before
+the file is erased.
 
-**Path**: `DELETE /file/{uuid}`
+The file deletion API is a two part process. The first request returns the bundles
+that will be **logically deleted** as a side effect and a confirmation
+code. The second request must include the confirmation code in order to begin the
+deletion process. The request follows this Swagger document:
```
Bento007: s/Swagger document/API specification

Bento007 (commented):
LGTM

hannes-ucsc (commented):
Does this account for the need to some logical deletions to never be followed up by physical deletions?

Bento007:
@hannes-ucsc yes it does. You would use the `physical` flag in the Bundle Deletion API Request to signify if it should be physically or logically deleted.

hannes-ucsc (approved):

rfcs/text/0004-dss-deletion-process.md
```diff
@@ -147,75 +156,80 @@ deleted from the DSS. The restoration will apply across all replicas.
...
-|uuid| A RFC4122-compliant ID for the file.|
-|version|Timestamp of file creation in DSS_VERSION format.|
-|Confirmation Code| A code used to confirm a deletion operation.|
+|uuid| An RFC4122-compliant ID for the bundle.|
+|version|Timestamp of bundle creation in DSS_VERSION format.|
+|confirmation_code| A code used to confirm a deletion operation.|
 
 |Response Code| Description|
 |--------------|------------|
 |200| The deleted bundle and associated files can be restored. A confirmation code is returned to confirm the deletion in the second request.|
```
aaclan-ebi, line 192: to confirm the restoration

rfcs/text/0004-dss-deletion-process.md
```diff
@@ -407,10 +421,10 @@ data that was permanently deleted over the previous 24 hours.
 
 ### Drawbacks and Limitations [optional]
 
-* This introduces the ability to permanently destroy data from the DSS. 
+* This introduces the ability to permanently destroy data from the DSS.
 * There is no limit on the amount of data that can be rapidly permanently deleted.
 * Permanently deletion does not occur until after the grace period has elapsed. This could
-  be problematic if files need to be removed sooner. 
+  be problematic if files need to be removed sooner.
 * If two bundles share files and only one bundle is deleted, this causes the remaining bundle to be incomplete.
```
aaclan-ebi, line 429: Is this how tombstoning currently works? almost all bundles share common files (e.g. project, donor) If a bundle is deleted, does that mean all other bundles wouldn't contain those shared files?

Bento007: Yes, it works like a tree. If a bundle is physically deleted, then all files in that bundle will be physically deleted, and all bundles containing those files will be logically deleted.

Bento007: The logically deleted bundles containing physically deleted files can be resubmitted, minus the deleted files to fix the bundle.

samanehsan (approved):

rfcs/text/0004-dss-deletion-process.md
aaclan-ebi: Just to check if my understanding is correct, if i have the following bundles and files:

Bundle 1 : 
- Project
- Donor
- Specimen
- Cell Suspension
- Sequencing File 1.1, Sequencing File 1.2, Sequencing File 1.3

Bundle 2: 
- Project
- Donor
- Specimen
- Cell Suspension 2
- Sequencing File 2.1 , Sequencing File 2.2, Sequencing 2.3

Bundle 3 : 
- Project,
- Donor,
- Specimen,
-  Cell Suspension 3,
-  Sequencing File 3.1, Sequencing File 3.2, Sequencing 3.3

If I requested Bundle 1 to be tombstoned ( just to "logically delete"):
1. will all files get tombstoned?
2. will Bundle 2 and 3 be tombstoned as well? 

I don't know if this is a valid use case, but will it be possible for the user to delete Bundle 1 but not delete Project, Donor, Specimen?

rfcs/text/0004-dss-deletion-process.md
natanlao: Good catch, thanks!

rfcs/text/0004-dss-deletion-process.md
natanlao: > 1. will all files get tombstoned?

No - if a bundle is logically deleted, files are unaffected.

> 2. will Bundle 2 and 3 be tombstoned as well?

No - Bundle 2 and Bundle 3 would be unaffected by a logical deletion of Bundle 1.

> I don't know if this is a valid use case, but will it be possible for the user to delete Bundle 1 but not delete Project, Donor, Specimen?

I think this is a valid use case. This can be accomplished by logically deleting the bundle. This logical deletion would also occur were a file only present in Bundle 1 physically deleted.

aaclan-ebi: I see, thanks!

aaclan-ebi (approved):

PR #119, opened 2019-09-25
RFC: HCA DCP Domain Oriented Bundle Types and Definitions

NoopDog:
**October  9**: Last call for community review.

Link to document view: [here](https://github.com/HumanCellAtlas/dcp-community/blob/76b4ee1b5cf8001d0437eec33647b7fd033edb8d/rfcs/text/0000-rfc-domain-oriented-bundle-types-and-definitions.md)

This RFC is offered as a refinement and generalization of the [RFC: HCA DCP Bundle Types and Definitions]( https://github.com/HumanCellAtlas/dcp-community/pull/93). 

This RFC describes:

1. An application/metadata level mechanism to define bundle types and a simple bundle type hierarchy.

1. An application/metadata level method to specify the type of a bundle.
 
1. Several bundle types designed to match current usage and near term expected usage. 

### Key Differences from RFC: HCA DCP Bundle Types and Definitions

This proposal differs from [RFC: HCA DCP Bundle Types and Definitions](https://github.com/HumanCellAtlas/dcp-community/pull/93/files) mainly in that:

1. Type information is added to a type.json  metadata file rather than added to the DSS bundle.json file.

1. Type information is expressed in JSON rather than of [RFC 7231](https://tools.ietf.org/html/rfc7231#section-3.1.1.1) media type syntax.

1. The schema and bundle types are represented as JSON schema in the metadata schema repo and documented on the Data Portal rather than maintained in a DSS registry.

1. Additional background and motivation is presented.


briandoconnor (approved):
Seems like a very good thing to nail down Dave.

Are the provenance links in the JSON DSS UUIDs and versions?  Maybe make that more clear?

There should be a plan to document this on a DCP section of the portal.  It should be one of our key SOP docs for the project I think.

This needs to be kept up to date over time.  If we add new types they need to be documented.

NoopDog:
> Are the provenance links in the JSON DSS UUIDs and versions? Maybe make that more clear?

Yes `provenance.document_id` is the type.json file DSS UUID. It's not clear to me exactly what the `provenance.submission_date` means in this context and if this maps reliably to the DSS file version in general. 

I have updated the doc to reflect the above.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,424 @@
...
+In practice, bundles are used to store the output of biological assays or analysis runs and their metadata although other uses are envisioned.
+
+Clients of the DSS register queries with the DSS to subscribe to notifications of new bundles matching the query. The queries are over the existence and contents of the metadata files making for a kind of “Duck Typing” as there are no explicit defined types. 
+
+### Rational for Bundle Types
+
+#### Subscription queries can become complicated and brittle 
+As the DSS is schemaless and there is no type system for bundles defined in the client application domain, subscription queries can become complicated as they must define selection criteria adequate to tell one assay or analysis type from another by looking at domain object properties only.   
+
+Additionally, there are on the horizon, requests to create bundles that specify lifecycle events of domain objects or groups of bundles (e.g. DAPS bundle) and to store reference data used by analyses, or the full metadata graph of a project for example. Care must be taken that any new bundle types do not accidentally match existing queries.
```
diekhans: The *DAPS* terminology was dropped,  The signaling bundles are data groups.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,424 @@
...
+*Recommended format for Shepherds:*
+
+ `[Name](mailto:username@example.com)`
+
+## Motivation
+
+When dealing with schemaless data stores such as DSS, the responsibility for establishing the data schema and ensuring data validity becomes the responsibility of client applications writing data into the data store.
+
+If the application does not specify a way to define and implement types, it is possible for multiple “ad hoc” types or document shapes to evolve over time that are not well documented and are difficult to describe, validate, and query. 
+
```
diekhans: I don't understand the meaning of "application" in the context.  I am lost a reviewing a lot of this without know this.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,424 @@
...
+## Shepherd
+***Leave this blank.** This role is assigned by DCP PM to guide the **Author(s)** through the RFC process.*
+
+*Recommended format for Shepherds:*
+
+ `[Name](mailto:username@example.com)`
+
+## Motivation
+
+When dealing with schemaless data stores such as DSS, the responsibility for establishing the data schema and ensuring data validity becomes the responsibility of client applications writing data into the data store.
```
diekhans: I am unclear on what is meant by *schemaless data stores*?  Is this no JSON Schema (as used by the experimental metadata) or is this more conceptual, as in the DSS stores blobs???

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,424 @@
...
+
+#### Bundle  process and protocol information is represented differently in ingest vs. analysis bundles
+Information about the process and protocols used to create a file are in different places in the current ad hoc ingest and secondary analysis bundles increasing the difficulty of determining the bundle origin across the ad hoc bundle types. 
+
+Ingest bundles specify a `process_0.json` file to indicate the process and a `sequencing_protocol_0.json` file to indicate the sequencing protocol. 
+
+Analysis bundles use an `analysis_process0.json` and `analysis_protocol_0.json`. Additionally because of copy forward, the `process_0.json` and `sequencing_protocol_0.json` also exist in the analysis bundles. 
+
+#### The bundle concept is not well understood by stakeholders
+One additional (and original) motivator for this RFC is that many DCP members, developers, data contributors and data consumers have trouble relating to the concept of a bundle. There are several factors that contribute to this:
```
diekhans: I think a big part of the problem is the lack of an abstraction layer between the experimental metadata/data and the underlying storage model.   We want to program against files, not bundles.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,424 @@
...
+
+1. For developers and DCP members, the difficulty may stem from the fact that we do not have an official shared vocabulary for referring to the entity type represented by the bundle.
```
diekhans: We need a terminology RFC!!

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,424 @@
...
+
+Analysis bundles use an `analysis_process0.json` and `analysis_protocol_0.json`. Additionally because of copy forward, the `process_0.json` and `sequencing_protocol_0.json` also exist in the analysis bundles. 
+
+#### The bundle concept is not well understood by stakeholders
+One additional (and original) motivator for this RFC is that many DCP members, developers, data contributors and data consumers have trouble relating to the concept of a bundle. There are several factors that contribute to this:
+
+1. For developers and DCP members, the difficulty may stem from the fact that we do not have an official shared vocabulary for referring to the entity type represented by the bundle. 
+
+1. Data contributors and consumers expect that their domain conceptual model maps rather directly to what they see in the DCP. Users want to search, download, and reference projects, assays, and analyses. They want the results of sequencing runs, alignment files, and expression counts. Storage level concepts such as the bundle and bundle.json are unexpected.
```
diekhans: bundle.json has information about files that is needed by the user (checksums, sizes, version).  It is also not possible for a developer to interpret version information without bundle.json.  It can't be hidden yet.

There is a problem in changing in of this in the lack of good layering between developer and storage layer.  It can be changed in one jump, it has to be eased in or it will fail

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,424 @@
...
+
+1. Bundle shapes vary widely by library construction approaches. For example a Smart Seq 2 ingest bundle contains a single cell, while a 10x V2 bundle may contain data for hundreds of cells. The fact that the ingest bundling size leaks all the way to the system output further complicates bundle comprehension. 
+
+1. Bundle are not well documented and are not self documenting.
```
diekhans: I don't think anything can truly be self-documenting

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,424 @@
...
+
+1. An application/metadata level mechanism to define bundle types and a simple bundle type hierarchy.
+
+1. An application/metadata level method to specify the type of a bundle.
+ 
+1. Several bundle types designed to match current usage and near term expected usage. 
+
+### Key Differences from RFC: HCA DCP Bundle Types and Definitions
+
+This proposal differs from [RFC: HCA DCP Bundle Types and Definitions](https://github.com/HumanCellAtlas/dcp-community/pull/93/files) mainly in that:
```
diekhans: I see this as a disadvantage.  The number of JSON objects required to actually understand the bundle is already large.  From a client perspective, not including type information in the manifest doesn't seem intuitive.   While the DSS strives to not be tied to just one project, this is an implementation detail we should not inflict on users.  Even if the generic DSS doesn't have a schema associated with bundle.json, the DCP can add one, which the schema link being a DSS configuration option.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,424 @@
...
+
+1. Type information is added to a type.json  metadata file rather than added to the DSS bundle.json file.
+
+1. Type information is expressed in JSON rather than of [RFC 7231](https://tools.ietf.org/html/rfc7231#section-3.1.1.1) media type syntax.
```
diekhans: Now that you mention it, this makes a lot of sense

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,424 @@
...
+
+1. The schema and bundle types are represented as JSON schema in the metadata schema repo and documented on the Data Portal rather than maintained in a DSS registry.
```
diekhans: As mention above, I don't think it is necessary to have a separate file so that we can have a schema.  Also, the current schema metadata repo is experimental metadata, so we need to somehow separate them, although I don't think putting them into different places serves the needs of the user, only distinguishing how bundle metadata and scientific metadata are not the same things.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,424 @@
...
+
+1. Additional background and motivation is presented.
```
diekhans: Always welcome so we don't continue to discuss the same thing over and over again.

diekhans (changes requested):
Only partway.  Generally, I really like the approach, but have outlined some concerns. 

Also, I worry this may be trying to do too much in one step and at the same time not enough to have an easier to use system.  Large changes are impossible to make without break the DCP.     We have to take steps, even if temporary, to get to the right place.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
NoopDog: Hi @diekhans I mean "application" as the client of the database as in web application, application server, service layer etc. Whoever writes into the database is responsible for maintaining data integrity and conformance to any desired schema as the database has no mechanism internally to validate the shape of the data.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,431 @@
+### DCP PR:
+
+Link to document view: [here](https://github.com/HumanCellAtlas/dcp-community/blob/d2da58b16efd80fa3e31ed2194169efd79b02aa2/rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md)
+
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC number and this pull request in the format:*
+
+`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+
+# Application Layer Bundle Types and Definitions
```
diekhans: I find the title of this RFC confusing, as I don't know what the application layer and have not seen it as part of any other DCP documentation.  Is this the data browser?   the API?

I think it would be better to not bring in terminology that isn't currently part of the mental DCP lexicon.

Although, we do need a written DCP lexicon!!

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
NoopDog: @diekhans I replaced DAPS with data_groups

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,431 @@
...
+
+
+## Summary
+
+This RFC is offered as a refinement and generization of the [RFC: HCA DCP Bundle Types and Definitions]( https://github.com/HumanCellAtlas/dcp-community/pull/93).
```
theathorn: generization -> generalization

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,431 @@
...
+
+1. Proposes a method of defining application level data types for DSS bundles.
+
+
+1. Proposes a method of implementing application level data types for DSS bundles.
+
+
+1. Proposes a simple type hierarchy for bundles that is closely tied to the application domain. 
+
+As the proposed updates to the bundle schema are at the application/metadata layer rather than the DSS/storage layer, no changes to DSS are envisioned, except for perhaps defining a new HCA file type and adding its contents to its index
```
theathorn: "new HCA file type". Do you mean bundle type?

NoopDog: I meant file as in the type.json file. I can make that more explicit.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
diekhans: Reading wikipedia: [Application layer](https://www.wikiwand.com/en/Application_layer), the title makes even less sense.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,431 @@
...
+
+1. Donor - similarly, a donor entity bundle could hold the entire metadata graph for a given donor in multiple formats.
+
+#### Event Bundle 
+
+Representing state change of a set of bundles relating to a domain entity or an arbitrary set of bundles. This could be used for example to implement a “data group” indicating that all of the Assay Bundles corresponding to a specific library prep instance have been ingested and secondary analysis on this “version complete” data set can commence. 
+
+### Example JSON Files
+
+#### Assay Bundle
```
diekhans: Merging links.json, which is part of the experimental metadata with bundle types defeats the goal of the metadata group at separating the metadata layer from the bundle layer.

NoopDog: The goal here is to tag the bundles with the process and protocol that created them in a uniform manner without disrupting the current system. Links.json use different file names for analysis vs assay bundles for the process and protocol links. Here the process and protocol are normalized across bundle types. 

Also by using the protocol and process id this gives the ultimate in type refinement on what is in this bundle. 

I do not see this as breaking a separation of concerns. The ids of the process and protocol used to create the bundle should not be uniquely visible to the "experimental metadata".  Our "processing model" as well should be able to benefit from the terms created in the experimental model to the extent that it is useful.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
NoopDog: I have added "What repo should the schema go in?" to the unresolved questions.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
NoopDog: I mean "Application Layer" in the multi tier architecture sense https://en.wikipedia.org/wiki/Multitier_architecture

Not the OSI Layer 7 networking protocol layer sense.

Screen Shot 2019-10-01 at 10 44 28 AM

NoopDog: I have added a note to the top of the doc that should clear the issue about the "Application Layer" term.

>Note: I have used the term "Application Layer" extensively in this document and the term seems to be causing some confusion. Let me clarify that I mean "Application Layer in the Multi Tier Architecture sense as opposed to the "Data Layer". Here the "Appplications" are the boxes or clients of DSS. Whoever writes into the database (DSS) needs to validate the records (bundles) they create are appropriately shaped as the database (DSS) does not to this.

I will resolve all comments about the "Application Layer" term. Please reopen if you feel this is not clear.

diekhans: It is still very confusing as the terminology is not used in the DCP and I still don't know what software comprises the application layer.  Azul?

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
diekhans: It the processes and protocols being referenced are metadata instances, this is very confusing

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
diekhans: I still feel that "application layer" is just introducing confusion from another programming model that is not used in the DCP.  "DSS client" or "DSS application" is clearer in reference to the DSS.  In terms of the overall DCP,

Maybe this is something to bring up in tech arch???

diekhans: p.s.  I do see the distinction trying to be made here and I think it is important, it is only the term that seems problematic.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
diekhans: > Yes `provenance.document_id` is the type.json file DSS UUID. It's not clear to me exactly what the `provenance.submission_date` means in this context and if this maps reliably to the DSS file version in general.
> 
> I have updated the doc to reflect the above.

Provenance section is added by ingest.  The submission_date may correspond to the ingest, however, is in a slightly different format so it should not be counted on. This is problematic and there is a ticket about this weirdness.

NoopDog: Also, this does not couple the experimental metadata with the bundle metadata, it couples the bundle metadata with a budding processing model in order to define the bundle type in _terms_ of the experimental metadata. This is a _use_, not a pollution of the experimental metadata.

diekhans: We can easily add new data to links.json too to address these issues.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,434 @@
...
+
+#### Bundle  process and protocol information is represented differently in ingest vs. analysis bundles
+Information about the process and protocols used to create a file are in different places in the current ad hoc ingest and secondary analysis bundles increasing the difficulty of determining the bundle origin across the ad hoc bundle types. 
+
+While internally the metadata is the same shape, the ingest bundles use a `process_0.json` file to indicate the process and a `sequencing_protocol_0.json` file to indicate the sequencing protocol. 
+
+Analysis bundles use an `analysis_process0.json` and `analysis_protocol_0.json`. Additionally because of copy forward, the `process_0.json` and `sequencing_protocol_0.json` also exist in the analysis bundles.
+
+Having the same information in different places depending on bundle type makes determing the process and protocol that created the bundle more difficlut than it needs to be.
```
TimothyTickle:
```suggestion
Having the same information in different places depending on bundle type makes determining the process and protocol that created the bundle more difficult than it needs to be. 
```

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
TimothyTickle: I know certain users will directly access the system but I feel that designing our internal storage based on user needs as opposed to abstracting or giving some presentation layer/interface could adversely affect our ability to flexibly engineer the backend. From what I have seen, users are perfectly happy and would want NFS-like file names used in object-store with metadata in the filenames or implied from the "directory" hierarchy but I would not make the DSS in this way. I would potentially deliver files in a tar.gz to certain users this way. Just a comment, no request for changes.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,434 @@
...
+
+1. “Imaging Assay” - representing the current imaging ingest bundles. 
+
+1. Other high level assay types.
+
+
+> _Note: It may take some discussion to decide on an effective set of subclasses. Sequencing assay may need to change to “2nd Gen Sequencing Assay” or “NGS Sequencing Assay” for example._
+
+#### Analysis Bundle
+Umbrella type used to represent an analysis with the following possible subclasses:
```
TimothyTickle: If the analysis outputs are going to be based on analysis types this will be a group with many subclasses as the diversity of analysis performed will only increase. No change requested, just FYI comment.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,434 @@
...
+
+
+1. “Alignment-Expression” - representing our current secondary analysis pipelines.
+
+
+1. “Expression Matrix” - a possible subtype representing the expression matrix generation.
```
TimothyTickle: This may simply be "matrix" given both count and expression matrices can currently exist. I believe in this case the shapes of the data would be the same but the measurements the matrices hold would be different. If this typing is how you would differentiate matrices that hold counts from matrices that hold expression it is important to note that they are not equivalent. In the case the subclassing is how those matrices are differentiated, I would recommend something like "count matrix" and "expression matrix" but once again these subtypes will balloon in number.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,434 @@
...
+
+
+1. “Expression Normalization” - a possible subtype representing the expression matrix normalization output for example.
+
+
+1. “Clustering” - a possible subtype representing cell clustering analyses output for example.
+
+> _Note: The exact names of the subclasses above can likely be improved upon. Please suggest._
```
TimothyTickle: Honestly, it may be good to just get the ones that exist standardized and then move forward as we extend the analysis. There are so many one can imagine we may need but we may not get to for a while.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,434 @@
...
+
+#### Reference Bundle
+Umbrella type representing a reference file used in analysis. Subclasses TBD.
```
TimothyTickle: I think we will have a better feel for this as we get more data modalities in the system. For now, this may be as simple as human and mouse. It may be good not to use the common name but use some specific ontology for the species names (I am sure our EBI friends have a favorite ontology to use). @lauraclarke

TimothyTickle (approved):
I have left some comments, most of which were cautionary and none of which should block this RFC. I feel the definition of bundles should really serve our technical/implementation team and so believe they should determine the fate of the RFC.

rfcs/text/0000-rfc-application-layer-bundle-types-and-definitions.md
NoopDog: Yes, these were meant to be suggestive. We would only define what we need to get going.

rfcs/text/0000-rfc-domain-oriented-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,362 @@
...
+This proposal differs from [RFC: HCA DCP Bundle Types and Definitions](https://github.com/HumanCellAtlas/dcp-community/pull/93/files) mainly in that:
+
+
+1. Type information is expressed in [JSON Schema](https://json-schema.org/) rather than of [RFC 7231](https://tools.ietf.org/html/rfc7231#section-3.1.1.1) media type syntax.
+
+1. An attempt is made to harmonize bundle types and properties with existing entities and field value sets defined in the scientific data model schema.
+
+1. The data_groups concept from [Processing Datasets that Span Multiple Data Collection Runs RFC](https://github.com/HumanCellAtlas/dcp-community/pull/88) is integrated and a corresponding bundle type is specified.
+
+1. The bundle types schema is maintained new ```bundle_schema``` Github repository and documented on the Data Portal rather than maintained in a DSS registry.
```
theathorn, line 33: grammar "maintained new"?

rfcs/text/0000-rfc-domain-oriented-bundle-types-and-definitions.md
```diff
@@ -0,0 +1,362 @@
...
+
+1. The bundle.json created by the HCA CLI download is renamed to manifest.json so that a new bundle.json can be used as a "bundle_descriptor" specifying the type information as a nomral file in the bundle.
```
theathorn, line 35: nomral -> normal

NoopDog:
Pausing to align with upcoming reproducibility and data citation requirements.

NoopDog:
Pausing to align with upcoming reproducibility and data citation requirements.

NoopDog:
Pausing to align with upcoming reproducibility and data citation requirements.

NoopDog:
Obsoleted

PR #129, opened 2019-11-19
Matrix Service RFC

calvinnhieu:
### Community Review
- [x] Add Labels
  - *For software RFCs*: add the "rfc-community-review" and the appropriate software project name (such as "Data 
  Store") labels. When the RFC impacts multiple DCP software projects, then the "Architecture" label MUST be the 
  project name. If uncertain about the appropriate project name, then Ask a PM on the 
  [HCA #dcp-project-mgmt](https://humancellatlas.slack.com/messages/C53S0GBS7) slack channel
  - *For governance RFCs*: Add the "rfc-community-review" and the "Governance" labels.
- [x] **Dec 3**: Community Review Last Call 
- [x] Request review on Slack channel [#dcp](https://humancellatlas.slack.com/messages/C51A8EDQQ).
- [x] Send an FYI to the appropriate HumanCellAtlas Slack channel. Include a link to the RFC pull request and the last call deadline.
  - For *Software RFCs*: [#tech-architecture](https://humancellatlas.slack.com/messages/C6SGU0R3J)
  - For *Governance RFCs*: [#dcp-project-mgmt](https://humancellatlas.slack.com/messages/C53S0GBS7)
- [x] **Shepherd**: Add RFC as an agenda item to next meeting:
  - For *Software RFCS*: DCP Architecture meeting
  - For *Governance RFCs*: DCP PM meeting
- [ ] **Shepherd**: Summarize the review discussion for the Approvers in the top-level summary comment in the RFC pull 
request  
  
### Approval
- [ ] Replace "rfc-community-review" with "rfc-approved"
- [ ] Rename the RFC from 0000-my-feature.md to ####-my-feature.md (with leading zeros) where #### is the next 
 available RFC number
- [ ] Create a link between the approved RFC and its pull request by updating the DCP PR section in the RFC template 
 and pushing the commit.

### Finally
Once all tasks are complete merge the PR in and Announce the approval in 
[#dcp](https://humancellatlas.slack.com/messages/C51A8EDQQ).


rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,408 @@
...
+
+*Related tables:* `analysis`, `donor`, `specimen`, `cell_suspension`, `library_preparation`, `project`, `publication`, `contributor`
+
+The majority of metadata is parsed directly from metadata JSON files available in primary bundles. These fields are identifiable as fields with a valid _Canonical Metadata Names_ mapping in _Table 2.1_. The exceptions are "derived fields" such as `derived_organ_*` that require a degree of intellectual reasoning of the experiment to determine an accurate value. For example, the value of `derived_organ_ontology` may differ from the value of `organ_ontology`, which is read directly from `specimen_from_organism.organ.ontology`, as it is determined by the presence of organoids and/or cell lines in the experiment.
+
+Metadata entities are identified by their `*.provenance.document_id` (DIDs) which are consistently used as primary keys in the schema defined in *Table 2.1*. A similar set of assumptions applied to _Cell IDs_ described in *Table 2.2* apply to these IDs as well:
+
+1. DIDs are unique for distinct metadata entities.
+2. DIDs are preserved across all bundles updates to preserve referential integrity of metadata entities over time.
```
justincc, line 332: The current plan for AUDR Analysis Updates will violate this assumption. On an analysis update, all the files will get different UUIDs apart from `analysis_process_0.json`.

calvinnhieu: Resolved in [Slack](https://humancellatlas.slack.com/archives/C6SGU0R3J/p1574189870221900)

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+
+The Matrix Service also does not have a migration strategy for its index and requires a full re-index to respond to metadata schema changes.
+
+#### Delete/Redact
+
+Deletion and redaction are handled the same way and are straightforward processes for the Matrix Service:
+
+1. Remove all data from Redshift that is associated with bundle FQIDs from a given project UUID
+2. Delete project matrices from Data Brower's S3 bucket
+3. Clear the Matrix cache (see [Availability and Reproducibility](#availability-and-reproducibility) for details on the cache)
```
brianraymor: It looks like *Availability and Reproducibility* were decomposed into separate sections which broke the anchor?

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+
+### Availability
+
+#### Request ID
+
+The `request_id` returned by `POST /matrix` is valid for 30 days following the successful completion of the request. Specifically, this refers to the S3 download URL to retrieve the generated matrix will be valid for 30 days. Following this time, matrices are deleted from S3 and `GET /matrix/request_id` will notify clients that the request has expired.
+
+#### Matrix cache
+
+Matrices are cached in S3 according to a hash of a [request's input parameters](#post-/matrix). Specifically, the hash encodes the set of `cellkey`s selected by the user's `filter` object, the list of metadata `fields` exported, the output `format`, and `feature`. During a request, matching hashes short-circuit the matrix generation process, a new copy of the existing matrix is created for this request, and the request is completed. This enables constant time delivery with respect to matrix size for cache hits compared to linear time for cache misses.
```
brianraymor: This should be `#post-matrix`.  github downcases, removes all punctuation, and then replaces ' ' with '-'.

rfcs/text/0000-matrix-service.md
calvinnhieu: Yes. Thanks!

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+
+#### Update
+
+The Matrix Service does not respond to bundle updates in real-time as the service does not subscribe to DSS notifications. This is primarily due to the Matrix cache and the loss in performance when invalidating entries when the index is updated. The cache provides constant time delivery of matrices with respect to matrix size for previously requested matrices, however, not responding to bundle updates risks data consistency across DCP services and data availability to users.
+
+To balance these competing motivations, the proposed SOP being discussed with the DataOps team is that bundle updates will occur on a monthly basis in which the Matrix Service will respond to on the same cadence. While this poses a limitation on the rate of data updates through the system, the team proposes that the Matrix Service continue this pattern and therefore guarantee support for DCP-wide data consistency for "Release Views" but not "Live Views".
```
calvinnhieu: This was discussed between myself, @mckinsel and @jlzamanian in 10/19. @jahilton or @maniarathi Is there a more formalized SOP on frequency of metadata updates?

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+
+### User Stories
+
+As a data consumer, I would like to access gene expression data from the DCP so that I can perform downstream integrative analysis across HCA experiments.
+
+- I would like to be able to receive an expression matrix comprised of cells that satisfy a set of metadata conditions.
+- I would like to receive metadata providing the experimental and scientific context about the cells in my matrix.
+- I would like to receive my matrix in a format that is compatible with standard single-cell analysis tools.
+
+## Scientific "guardrails"
```
brianraymor, line 39: Consider adding guidance related to multiple species:


[**August 6 - Guidance from AWG**](https://docs.google.com/document/d/1YcSg2DhePi1Ozb0ty7_pGdAPiPeN9B6WaDhuN0PiYRU/edit#heading=h.qf4cth7gazsu)

Consider a "give me all kidney cells" query: 
- Only return human kidney cells, mouse requires an explicit "from mouse" in the query
  - This assumes all queries are human by default.
  - This is the requested default for a query not specifying species.

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC number and this pull request in the format:*
+
+`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+
+# Matrix Service
+
+## Summary
+
+The existing Matrix Service is a prototype of a DCP product that is capable of enabling third party data consumers to perform integrative analysis across HCA experiments. In its current implementation, this is achieved by 1) delivering a RESTful API that enables data consumers to generate expression matrices defined by a metadata query against an index of cells, and 2) an index of cells in the HCA that belong to projects for which a secondary analysis pipeline is available and complete. This RFC details and requests community review on the prototype's major design decisions surrounding i) the Matrix Service API and ii) the service's cell-centric data model.
```
maniarathi: - The Matrix Service isn't really a DCP _product_ per se but a feature? Functionality? Something like that.
- "is capable of enabling" -> psh. "enables". Assertive language :)
- "enabling third party data consumers" -> don't need to refer to data consumers as "third party" so you can probably just delete those two words.
- "across HCA experiments" -> I think the beauty is actually across _projects_, thus providing the integrative analysis.
- The indexing is more of an implementation detail, no? The integrative analysis is achieved by the delivery of the RESTful API that generates the matrices. If so, you could omit it here.
- Can delete "requests community review"; one day this will be approved and then it won't be requesting community review any longer.

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+
+## Authors
+
+[Calvin Nhieu](cnhieu@chanzuckerberg.com)
+[Marcus Kinsella](mkinsella@chanzuckerberg.com)
```
maniarathi, line 17: Need an extra line break here so that your names are on separate lines.

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+
+## Shepherd
+ [Brian Raymor](braymor@chanzuckerberg.com)
+
+## Motivation
+
+The Human Cell Atlas (HCA) Data Coordination Platform (DCP) ingests, processes, and stores single-cell data, making that data easily available to a wide audience of researchers. The single-cell data type of greatest interest to most researchers is the expression matrix, the cell-by-gene expression values that are the starting point for many downstream analyses. To support this most commonly requested data type, the DCP offers a specialized interface, the DCP matrix service, which enables investigators to easily access and perform integrative analysis across HCA expression data.
```
maniarathi: - The first occurrence of HCA DCP is in the Summary section so I'd pop the fully written out terms up there and then use HCA/DCP throughout. 
- Could elaborate a bit on "downstream analyses" -> "downsteam analysis such as cell type annotation, differential expression, etc."
- Nit: capitalize Matrix Service throughout to be consistent?

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+
+### User Stories
+
+As a data consumer, I would like to access gene expression data from the DCP so that I can perform downstream integrative analysis across HCA experiments.
```
maniarathi: - I think we still generally prefer using one of the UX-based user profiles? "As a user with a keyboard, I would like to... "
- s/"I would like to access gene expression.. "/"I would like gene expression.. "

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+
+- I would like to be able to receive an expression matrix comprised of cells that satisfy a set of metadata conditions.
```
maniarathi: s/receive/generate?

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+[Marcus Kinsella](mkinsella@chanzuckerberg.com)
+
+## Shepherd
+ [Brian Raymor](braymor@chanzuckerberg.com)
+
+## Motivation
+
+The Human Cell Atlas (HCA) Data Coordination Platform (DCP) ingests, processes, and stores single-cell data, making that data easily available to a wide audience of researchers. The single-cell data type of greatest interest to most researchers is the expression matrix, the cell-by-gene expression values that are the starting point for many downstream analyses. To support this most commonly requested data type, the DCP offers a specialized interface, the DCP matrix service, which enables investigators to easily access and perform integrative analysis across HCA expression data.
+
+### User Stories
```
maniarathi, line 26: So far I haven't read anything about performance but I think one of the cool things about Matrix Service is that it doesn't take a bajillion years to generate these matrices. Might add a user story or something in the summary/motivation about "the need for speed."

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+
+## Detailed Design
+
+### Overview
+
+The Matrix Service enables on-the-fly expression matrix generation via a [RESTful API](https://matrix.data.humancellatlas.org/). Through user studies and iterative development over the past year, four major requirements of matrices have been identified and must be met. One, matrices must enable integrative analysis across data in the HCA and the Matrix Service determines that this integration should be at the granularity of individual cells. Second, matrices should include rich metadata that users are interested in so that they can better understand the data they're working with. Third, matrices should be compatible with downstream single cell analysis tools, for example in `Python` and `R`, to further enable downstream analysis of HCA data by the wider community. Finally, performance of serving matrices should be prioritized as the scale of HCA data is anticipated to reach or exceed the order of 10<sup>7</sup> cells. The Matrix Service API is designed to fulfill these requirements in order to provide access to HCA data in a format that enables and encourages downstream integrative analysis.
```
maniarathi: - Could you break out the 4 requirements into bullet points please?
- Matrix Service didn't determine that the integration should be at the cell-granularity.. it was the users, right? Could reword to "matrices must enable cell-level granularity integrative analysis across data in the HCA"
- On your second point, might swap the word "should" to "must" since it was a requirement.
- On your third point, "for example in Python and R" isn't very descriptive of the analysis tools. I think these could be "Seurat and Scanpy"? (@mckinsel to confirm)
- On your fourth point, instead of using 10^7, maybe just write out "tens of millions" for clarity.

rfcs/text/0000-matrix-service.md
maniarathi: - Oh boy. Have we formally defined "Release View" vs "Live View" yet? I think there's a difference between "signed out" and "signed in" users but I'd refrain from using particular terms for now.
- No formalized SOP for metadata updates for now. Currently they can be pushed whenever and components are free to accommodate the metadata changes on a cadence they see fit. This will change as we start to enforce interoperability.

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+- [Motivation Aug 2018](https://github.com/HumanCellAtlas/matrix-service/blob/d797cc96c645f90df8fcb4bb9252074931ecd108/MOTIVATION.md)
+- [Use Cases Feb 2018](https://docs.google.com/document/d/12gfF3UEgRZ0YJiBoScLMtcW6Cthv0_MZHQwj4WMYKPA/edit#)
+- [API v0](https://github.com/HumanCellAtlas/matrix-service/blob/1c43b9f3d757ffcb9111b5a13d46691249435f36/README.md)
+  - Issues:
+    - Read expression data directly from DSS which posed significant performance limitations
+    - Exposed bundles via API which caused confusion and data access limitations
+    - Metadata was not exported with matrices
+    - `zarr` previously supported as default
+
+### Alternatives
```
maniarathi: Can delete this section since empty.

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+
+A discussion should be had with the DCP regarding the feasibility of DIDs meeting these assumptions. (1) is potentially at risk as it is possible for the same metadata entity to have different DIDs in separate projects. (2) may be true and requires confirmation from Wranglers/Ingest. Work to satisfy (3) is ongoing and tracked [here](https://github.com/HumanCellAtlas/ingest-central/issues/598).
+
+#### Feature data
+
+*Populated tables:* `feature`
+
+Feature data is not parsed from files in primary or analysis bundles. Rather, they are retrieved from GENCODE and specifically, from the primary annotation GTF matching the version used by Secondary Analysis, i.e. [v27 for human](https://www.gencodegenes.org/human/release_27.html). The ETL job will download the GTF and parse gene and transcript data to populate the `feature` table. There currently is no process in place for synchronizing GTF versions between Secondary Analysis and Matrix Service.
+
+### AUDR
```
maniarathi, line 343: Probably good to define what AUDR is here.

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+
+#### Metadata
+
+*Related tables:* `analysis`, `donor`, `specimen`, `cell_suspension`, `library_preparation`, `project`, `publication`, `contributor`
+
+The majority of metadata is parsed directly from metadata JSON files available in primary bundles. These fields are identifiable as fields with a valid _Canonical Metadata Names_ mapping in _Table 2.1_. The exceptions are "derived fields" such as `derived_organ_*` that require a degree of intellectual reasoning of the experiment to determine an accurate value. For example, the value of `derived_organ_ontology` may differ from the value of `organ_ontology`, which is read directly from `specimen_from_organism.organ.ontology`, as it is determined by the presence of organoids and/or cell lines in the experiment.
```
maniarathi: Do you parse secondary analysis metadata as well?

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+
+Metadata entities are identified by their `*.provenance.document_id` (DIDs) which are consistently used as primary keys in the schema defined in *Table 2.1*. A similar set of assumptions applied to _Cell IDs_ described in *Table 2.2* apply to these IDs as well:
+
+1. DIDs are unique for distinct metadata entities.
+2. DIDs are preserved across all bundles updates to preserve referential integrity of metadata entities over time.
+3. DIDs are versioned so that metadata can be synchronized across DCP services.
+
+A discussion should be had with the DCP regarding the feasibility of DIDs meeting these assumptions. (1) is potentially at risk as it is possible for the same metadata entity to have different DIDs in separate projects. (2) may be true and requires confirmation from Wranglers/Ingest. Work to satisfy (3) is ongoing and tracked [here](https://github.com/HumanCellAtlas/ingest-central/issues/598).
```
maniarathi, line 335: Could pop this into the Unresolved Questions section.

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+{
+    "field_description": "Count of genes with a non-zero count.",
+    "field_name": "genes_detected",
+    "field_type": "numeric",
+    "maximum": 13108,
+    "minimum": 2
+}
+```
+
+### Cell-centric Index
```
maniarathi, line 282: Somewhere in here, I'd suggest having a diagram to visualize the architecture. Helpful to associate with the descriptions.

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+
+#### Format
+
+The API generates expression matrices in three formats: `loom`, `mtx` and `csv`. From the [loom](http://loompy.org/) documentation page, the format is "an efficient file format for large omics datasets", enabling the storage of large expression matrices with metadata and also supports many popular data processing programming languages. For these reasons, `loom` is the service's default and most performant format. The `mtx` format adheres to the [10x feature-barcode matrix specification](https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/output/matrices) to enable compatibility with `scanpy` and `seurat` 10X methods. `csv` is supported as a raw text format for which performance may not be optimized for.
```
maniarathi: - The loom description is already written above so could delete here? (line 43)
- scanpy and seurat aren't 10x methods? 
- can delete "for" in the last sentence.

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+	'filter': dict,
+	'fields': list,
+	'format': string,
+	'feature': string
+}
+```
+
+#### Filter
+
+The filter object is inspired by the [GDC Search and Retrieval API](https://docs.gdc.cancer.gov/API/Users_Guide/Search_and_Retrieval/#filters-specifying-the-query). To enable users the ability to express complex metadata queries, the object supports nested composition of base filter objects to create a tree of AND/OR statements. The API deviates from the GDC API specification by defining two types of base filter objects:
```
maniarathi: What's GDC?

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,406 @@
...
+
+The default and recommended file format for matrices generated by the service is [loom](http://loompy.org/), an HDF5 file format that is optimized for storing large omics datasets such as expression matrices and designed for consumability, for example, by exporting metadata with matrices. The Matrix Service achieves its first three goals by delivering matrices in `loom`, and other supporting formats, via an expressive API that enables users the ability to specify individual cells and desired metadata to generate a matrix for. For a detailed example, see [this user vignette](https://nbviewer.jupyter.org/github/HumanCellAtlas/matrix-service/blob/master/docs/HCA%20Matrix%20Service%20to%20Scanpy.ipynb#HCA-Matrix-Service-Loom-Output) describing a `loom` matrix generated by the service. The fourth requirement is satisfied by a lazy cache of matrices providing constant time delivery of previously requested matrices.
```
maniarathi: How fast is "constant time"? I think this is a very computer science specific term so maybe add some concreteness to it?

maniarathi: I see three locations where "constant time" is mentioned so maybe correct throughout?

maniarathi (changes requested):

rfcs/text/0000-matrix-service.md
calvinnhieu: - 👍 
- 👍 
- 👍 
- Intended to use the words interchangeably - will opt for "projects"
- Does "here" refer to that specific sentence in the Summary? If so, agreed - omitting
- 👍

rfcs/text/0000-matrix-service.md
calvinnhieu: Ah, you're right we do. Good catch!

@justincc From our [earlier conversation](https://humancellatlas.slack.com/archives/C6SGU0R3J/p1574188428221000), you mentioned only `document_id`s  from `analysis_process_0.json` will be preserved on analysis bundle updates, correct? Our index stores `analysis_protocol`s and uses its `document_id` as a primary key. It sounds like the current AUDR analysis updates plan will in fact violate our assumptions on DIDs.

My question is, is it possible for these DIDs to be generated deterministically?

rfcs/text/0000-matrix-service.md
calvinnhieu: - Would prefer to keep the section self-sufficient
- Refers to 10x methods from `scanpy` and `seurat`. Will reword
- :+1:

rfcs/text/0000-matrix-service.md
calvinnhieu: Genomic Data Commons. Will expand

rfcs/text/0000-matrix-service.md
calvinnhieu: - Agreed. Sorry I was loosely trying to describe what our position was. Rewriting.
- Good to know. Will reflect in the update.

rfcs/text/0000-matrix-service.md
calvinnhieu: Great idea!

rfcs/text/0000-matrix-service.md
calvinnhieu: Sure, will link [arch doc](https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/matrix-service/wikis/Technical-Architecture) somewhere, however, the architecture itself feels overly technically bloated wrt to the RFC so I'd opt to not include it directly. My hope is the comprehensive table schema provides a sufficient picture of the index.

rfcs/text/0000-matrix-service.md
calvinnhieu: To answer your question, minutes if not less. Good call out, will update

rfcs/text/0000-matrix-service.md
justincc: The current plan is that submitters (Analysis) will be able to specify the UUID. So it would be up to them to generate a deterministic protocol UUID as well as one for the analysis process. @samanehsan and @rexwangcc could comment further.

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,413 @@
...
+The HCA DCP ingests, processes, and stores single-cell data, making that data easily available to a wide audience of researchers. The single-cell data type of greatest interest to most researchers is the expression matrix, the cell-by-gene expression values that are the starting point for many downstream analyses (e.g. cell type annotation, differential expression, etc.). To support this most commonly requested data type, the DCP offers a specialized interface, the DCP Matrix Service, which enables investigators to easily access and perform integrative analysis across HCA expression data.
+
+### User Stories
+
+As a user with a keyboard, I would like gene expression data from the DCP so that I can perform downstream integrative analysis across HCA experiments.
+
+- I would like to generate an expression matrix comprised of cells that satisfy a set of metadata conditions.
+- I would like to generate metadata providing the experimental and scientific context about the cells in my matrix.
+- I would like to generate my matrix in a format that is compatible with standard single-cell analysis tools.
+- I would like to generate my matrix in a reasonable amount of time (< 24 hrs).
```
jahilton: I'm betting a user's definition of 'reasonable amount of time' for this is much closer to 1 min than 1 day.

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,413 @@
...
+## Shepherd
+ [Brian Raymor](braymor@chanzuckerberg.com)
+
+## Motivation
+
+The HCA DCP ingests, processes, and stores single-cell data, making that data easily available to a wide audience of researchers. The single-cell data type of greatest interest to most researchers is the expression matrix, the cell-by-gene expression values that are the starting point for many downstream analyses (e.g. cell type annotation, differential expression, etc.). To support this most commonly requested data type, the DCP offers a specialized interface, the DCP Matrix Service, which enables investigators to easily access and perform integrative analysis across HCA expression data.
+
+### User Stories
+
+As a user with a keyboard, I would like gene expression data from the DCP so that I can perform downstream integrative analysis across HCA experiments.
```
jahilton: How does the Matrix Service view the needs of a user without a pipette (or a user with a keyboard that wants to poke around for a bit before going to their keyboard)? Are their needs met by a different component that puts a layer between the MS API & the user?

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,413 @@
...
+
+#### Feature
+
+Users are also able to specify the expression feature type, i.e. the matrix's axis label, as `gene` (default) or `transcript` names.
+
+#### Species (proposed)
+
+`species` is a potential fifth parameter that the endpoint may accept. Currently, `species` is not a top-level parameter and is rather specified within `filter` via the `cell_suspension.genus_species.ontology` and `cell_suspension.genus_species.ontology_label` filters. Since a distinct matrix is generated for each `species` requested, it may be a more intuitive user experience for it to be a top-level parameter. Additionally, it requires non-trivial logic for the Matrix Service to parse a complex `filter` object to determine which species are requested, i.e. which species to generate a matrix for. In order to simplify the user's mental model of the API as well as the service's logic complexity, `species` is proposed as a fifth request body parameter.
+
+*Note: Currently, supported species are Homo Sapiens and Mus Musculus.*
```
jahilton:
```suggestion
*Note: Currently, supported species are _Homo sapiens_ and _Mus musculus_.*
```

rfcs/text/0000-matrix-service.md
calvinnhieu: Agreed... The 1 day was being generously inclusive (least performant format, uncached, full HCA request). Will update to reflect majority use case as opposed to worst case.

rfcs/text/0000-matrix-service.md
calvinnhieu: *with a pipette? 

Matrices are made available per project, for all formats, via the Data Browser which would be the most accessible point for users with a pipette. This seems more like a Data Browser requirement than one for MS though.

rfcs/text/0000-matrix-service.md
calvinnhieu: The note is already italicized - will only italicize species names.

jahilton: Sorry, didn't realize that. Would've been fine then.
Please lowercase the species names

rfcs/text/0000-matrix-service.md
jahilton: Yes, thanks for correcting that.
I think a note would be good explicitly stating MS's stance regarding the other major user group.

rfcs/text/0000-matrix-service.md
calvinnhieu: Added a sentence

rfcs/text/0000-matrix-service.md
calvinnhieu: Done

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,418 @@
...
+
+***Leave this blank until the RFC is approved** then the **Author(s)** must create a link between the assigned RFC number and this pull request in the format:*
+
+`[dcp-community/rfc#](https://github.com/HumanCellAtlas/dcp-community/pull/<PR#>)`
+
+# Matrix Service
+
+## Summary
+
+The existing Matrix Service is a prototype of a Data Coordination Platform (DCP) feature that enables data consumers to perform integrative analysis across Human Cell Atlas (HCA) projects. In its current implementation, this is achieved by delivering a RESTful API that enables data consumers to generate expression matrices defined by a metadata query against an index of cells. This RFC details the prototype's major design decisions surrounding i) the Matrix Service API and ii) the service's cell-centric data model.
```
samanehsan, line 11: Why is the matrix service still considered a prototype feature? Should that be communicated to consumers?

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,418 @@
...
+| Comparison Filter | {<br />   `op`: one of `[ =, !=, >, <, >=, <=, in ]`,<br />   `field`: a metadata field,<br />   `value`: string \| int \| list <br />} |
+| Logical Filter    | {<br />   `op`: one of `[ and, or, not ]`,<br />   `value`: array of >=2 filter objects `if op==and|or`, filter object `if op==not` <br />} |
+
+Comparison Filters specify a condition on a metadata field and Logical Filters implement `and`, `or` and `not` logic. Note that the `value` field of a Logical Filter accepts a base filter object and is the mechanism for nesting chains of AND/OR statements. For a Comparison Filter's `field` parameter, the list of accepted metadata filters is available at the `/filters` endpoint. More information about a filter such as a description and a set of valid values can be found at the `/filters/<filter>` endpoint. See [Auxiliary Endpoints](#auxiliary-endpoints) for more details on these.
+
+Ex. Select all full length cells:
+
+```
+"filter": {
+  "op": ">=",
```
samanehsan: Is this meant to be "=" instead of ">="?

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,418 @@
...
+The response object returns a `request_id` that users can use to retrieve the status and results of a request by supplying it to the `GET` `/matrix` endpoint. A `status` and `message` field are also supplied to inform the user whether the request started successfully or not.
+
+*Note about multiple species:*
+If a request selects cells from multiple species, the Matrix Service will generate one matrix per species and thus return multiple `request_id`s. The convention here is for the human matrix to be reflected in `request_id` and all other species matrices to be reflected in a list of `request_id`s via `non_human_request_ids`.
+
+```
+{
+	'request_id': string,
+	'non_human_request_ids': list,
+	'status': one of ["In Progress", "Complete", "Failed", "Expired"],
```
samanehsan, line 169: When would a request have the `expired` status?

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,418 @@
...
+| expression          | cellkey (PK)<br />featurekey (FK)<br/>exprtype<br/>exprvalue      | N/A (derived field)<br />N/A (not metadata)<br />N/A (not metadata)<br />N/A (not metadata) | Stores gene expression values. Each row specifies a cell, feature (e.g. gene) and expression value. |
+| feature             | featurekey (PK)<br />featurename<br/>featuretype<br/>chromosome<br/>featurestart<br/>featureend<br/>isgene<br/>genus_species | N/A (not metadata)<br />N/A (not metadata)<br />N/A (not metadata)<br />N/A (not metadata)<br />N/A (not metadata)<br />N/A (not metadata)<br />N/A (not metadata)<br />cell_suspension.genus_species.ontology_label | Describes gene and transcript information for supported species. Data is populated from GENCODE GTF annotation files. |
+| analysis            | analysiskey (PK)<br />bundle_fqid<br/>bundle_uuid<br/>bundle_version<br/>protocol<br/>awg_disposition | analysis_protocol.provenance.document_id<br />dss_bundle_fqid<br />bundle_uuid<br />bundle_version<br />analysis_protocol.protocol_core.protocol_id<br />analysis_working_group_approval_status | Represents an analysis bundle.                               |
```
samanehsan, line 293: We do not provide a field for `analysis_working_group_approval_status` because there is no way to update that value when the status of a pipeline changes, so we are only running approved pipelines for now.

rfcs/text/0000-matrix-service.md
calvinnhieu: Hm the moniker is more so to communicate that the design is still up for review per this RFC. Any major API changes following review will be communicated to the community via #the-matrix-reloaded

rfcs/text/0000-matrix-service.md
calvinnhieu: Ah, yup! Good eye

rfcs/text/0000-matrix-service.md
calvinnhieu: Following the 30 day `request_id` validity period. This is mentioned in _Availability_, but I'll add a link from here to there for clarity!

rfcs/text/0000-matrix-service.md
calvinnhieu: :+1: we basically always [set this field](https://github.com/HumanCellAtlas/matrix-service/blob/master/matrix/common/etl/transformers/analysis.py#L32) to `blessed` for the time being.

samanehsan: I think the new term is "standard" instead of "blessed", right @kbergin?

kbergin: Yup, it is.

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,418 @@
...
+
+The default and recommended file format for matrices generated by the service is [loom](http://loompy.org/), an HDF5 file format that is optimized for storing large omics datasets such as expression matrices and designed for consumability, for example, by exporting metadata with matrices. The Matrix Service achieves its first three goals by delivering matrices in `loom`, and other supporting formats, via an expressive API that enables users the ability to specify individual cells and desired metadata to generate a matrix for. For a detailed example, see [this user vignette](https://nbviewer.jupyter.org/github/HumanCellAtlas/matrix-service/blob/master/docs/HCA%20Matrix%20Service%20to%20Scanpy.ipynb#HCA-Matrix-Service-Loom-Output) describing a `loom` matrix generated by the service. The fourth requirement is satisfied by a lazy cache of matrices that enables the availability of previously requested matrices within minutes, if not seconds.
+
+*Note: The current technical architecture of the service is documented [here](https://allspark.dev.data.humancellatlas.org/HumanCellAtlas/matrix-service/wikis/Technical-Architecture).*
+
```
lauraclarke, line 59: Is there a plan to put this documentation in a publicly accessible location. I don't have an allspark account and I can't read these documents?

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,418 @@
...
+#### Feature
+
+Users are also able to specify the expression feature type, i.e. the matrix's axis label, as `gene` (default) or `transcript` names.
+
+#### Species (proposed)
+
+`species` is a potential fifth parameter that the endpoint may accept. Currently, `species` is not a top-level parameter and is rather specified within `filter` via the `cell_suspension.genus_species.ontology` and `cell_suspension.genus_species.ontology_label` filters. Since a distinct matrix is generated for each `species` requested, it may be a more intuitive user experience for it to be a top-level parameter. Additionally, it requires non-trivial logic for the Matrix Service to parse a complex `filter` object to determine which species are requested, i.e. which species to generate a matrix for. In order to simplify the user's mental model of the API as well as the service's logic complexity, `species` is proposed as a fifth request body parameter.
+
+Note: Currently, supported species are _Homo Sapiens_ and _Mus Musculus_.
```
lauraclarke, line 156: How species specific is the matrix service? is this parameterization where by if someone wanted to setup a matrix service which served mouse first or even on say the fly cell atlas or other species as they are created that it could be acheived by changing parameters or is this a change which would require more significant software changes?

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,418 @@
...
+
+#### Response
+
+The response object returns a `request_id` that users can use to retrieve the status and results of a request by supplying it to the `GET` `/matrix` endpoint. A `status` and `message` field are also supplied to inform the user whether the request started successfully or not.
+
+*Note about multiple species:*
+If a request selects cells from multiple species, the Matrix Service will generate one matrix per species and thus return multiple `request_id`s. The convention here is for the human matrix to be reflected in `request_id` and all other species matrices to be reflected in a list of `request_id`s via `non_human_request_ids`.
+
```
lauraclarke, line 164: Do you have a usability testing plan for this? from a very naive standpoint that feels like a lot of additional work to get the data for non-human species? What was the rationale to not return an array of matrix files?

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,418 @@
...
+        "HumanTissueTcellActivation": 267360,
+        "KidneySingleCellAtlas": 341079,
+        "Reprogrammed_Dendritic_Cells": 15744,
+        "Single cell RNAseq characterization of cell types produced over time in an in vitro model of human inhibitory interneuron differentiation.": 1733,
+        "Single cell transcriptome analysis of human pancreas": 2544,
+        "SingleCellLiverLandscape": 299486,
+        "Tissue stability": 73515,
+        "WongAdultRetina": 41350
+    },
+    "field_description": "A short name for the project.",
```
lauraclarke, line 264: Are field descriptions where possible taken directly from the metadata schema?  Are there any field descriptions for fields defined in the schema which don't match the schema?

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,418 @@
...
+
+The team agrees to a 1-week processing period to make available a project following secondary analysis completion with the current manual process. Following automation, this period will be updated to 24 hours to index the project and 48 hours for availability via the Data Browser.
+
+#### Update
+
+The Matrix Service does not respond to bundle updates in real-time as the service does not subscribe to DSS notifications. This is primarily due to the [Matrix cache](#matrix-cache) and the loss in performance when invalidating entries when the index is updated. The cache dramatically expedites the time to receive a matrix, however, not responding to bundle updates risks data consistency across DCP services and data availability to users.
+
+To balance these competing motivations, the team proposes to respond to updates on a monthly cadence in the short term. In the long term, the team aims to support DCP-wide data consistency for versioned releases of HCA data. See [Reproducibility](#reproducibility) for more details on data guarantees.
```
lauraclarke, line 359: It would be good to understand what classes of updates that the matrix services are proposing we conduct on a monthly cadence.

The metadata team is aware that right now we have a higher rate of errors in our metadata that is ideal and this creates churn which impacts all DCP teams. Data Accuracy is a priority for Q1 and we will work with the Data Ops team to create a virtous circle so as new error modes are discovered we can as rapidly as possible update our validation to catch the types of errors that can be caught automatically. This means overtime we should see the rate of errors drop and the frequency by which updates are needed decrease

My concern with only having a monthly cadence for metadata updates is when we have discovered an error in a metadata field which about the impact the scientific interpretability of the data, the easiest of those to understand is if a sample has been labelled with the wrong biological sex. This is a class of error we will never be able to catch automatically but when we are notified of such an error I would want to ensure we are in a position where we can fix it in 24 to 48hrs not 1 month.

Can we work together to understand what the best way forward is here to ensure updates don't create challenges for consumer components but we also don't have to leave erroneous metadata in the system for longer than necessary.

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,418 @@
...
+    "bundle_version",
+    "file_uuid",
+    "file_version",
+    "analysis_protocol.protocol_core.protocol_id",
+    "analysis_working_group_approval_status"
+]
+```
+
+Ex. Categorical filter: GET /filters/project.project_core.project_short_name
+
```
jychien, line 246: The 'GET /filters/<filter>' is a great feature for users to get a better idea of how many cells will be retrieved from a matrix request before actually requesting it. Is there any way to combine the filters to get cell counts across multiple categories? For example, what if the user would like to know how many cells there are when categorized by 'derived_organ_label' and also 'library_preparation_protocol.library_construction_method.ontology_label'.

rfcs/text/0000-matrix-service.md
calvinnhieu: There isn't a plan to -  though, are you able to authenticate with GitHub? If not, I can see if we can get you access. Several DCP components host their developer documentation on GitLab (allspark)

rfcs/text/0000-matrix-service.md
calvinnhieu: Could you elaborate on "setup a matrix service"?

Loading the new species data into the Matrix Service requires some software changes such as supporting new genomes, but once the data is available users are able to generate a matrix for that species by specifying parameters in the POST endpoint.

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,418 @@
...
+Project matrices served via the Data Browser are regenerated and updated when projects are reingested or relevant bundle updates are applied to the index.
+
+### Reproducibility
+
+The Matrix Service will maintain a single version of a bundle for which will be the latest version that belongs to a [version-complete](https://github.com/HumanCellAtlas/dcp-community/blob/master/rfcs/text/0015-project-completion-stage.md#definitions) project. The following reproducibility implications apply:
+
+1. Matrices for historical versions of bundles/projects can not be generated.
+2. The Matrix Service is non-deterministic over time. Identical parameters supplied to the Matrix Service at two different times does not guarantee identical outputs. 
+3. The current Matrix Service does not support an implementation for Citations or Snapshots.
+
```
jychien, line 396: Agreed that the Matrix Service is non-deterministic and that reproducibility would be out of scope for Matrix Service. However, just from the perspective of a user that potentially will generate many matrices with multiple filters in a short amount of time, it is easy to get confused and tedious to organize as to which matrix corresponds with which post request. I remember this was brought up before, but would it possible to add logs (not to imply reproducibility but rather for historical logging purposes)?

rfcs/text/0000-matrix-service.md
calvinnhieu: As the API evolves, we should be explicit about testing new changes such as this one in our future User Studies.

This design choice was made for a couple reasons:

1. So as to not introduce a breaking API change that changing the type returned by `request_id` would
2. The initial design assumed 1 output matrix per 1 request ID and is reflected in design decisions throughout the software. Returning multiple matrices per request ID would compromise the design and introduce, in our opinion, unnecessary software complexity (e.g. many other 1:1 mappings in the code would need to become 1:N). This in addition to the project being the _Human Cell Atlas_ supported this cleaner, simpler implementation.

With that being said, the consideration for the user's experience is paramount and should this design prove to be inadequate for users retrieving non-human data, the team acknowledges that the software should accommodate the more conducive design.

rfcs/text/0000-matrix-service.md
calvinnhieu: I believe (1) this is true, and (2) no. @mckinsel can you confirm?

rfcs/text/0000-matrix-service.md
```diff
@@ -0,0 +1,418 @@
...
+    "barcode",
+    "emptydrops_is_cell",
+    "specimen_from_organism.provenance.document_id",
+    "cell_suspension.genus_species.ontology",
+    "cell_suspension.genus_species.ontology_label",
+    "donor_organism.provenance.document_id",
+    "donor_organism.human_specific.ethnicity.ontology",
+    "donor_organism.human_specific.ethnicity.ontology_label",
+    "donor_organism.diseases.ontology",
+    "donor_organism.diseases.ontology_label",
```
jychien, line 211: The field used to populate the browser for 'Disease Status' is specimen_from_organism.diseases.ontology_label. This is because there can be 'normal' specimen collected from donors with other non-related diseases. Would it be worthwhile to also include specimen disease ontology information?

rfcs/text/0000-matrix-service.md
calvinnhieu: Not without requesting the matrix. In your example, I imagine the number of combinations of field values could become unmanageable. In the future, possibly a similar API to `POST /matrix` (or enabled as a parameter on the existing endpoint) that retrieves cell counts without generating a matrix could be helpful. I'd be interested in seeing what the community interest for this would be. @mckinsel thoughts?

rfcs/text/0000-matrix-service.md
calvinnhieu: Yes, I've run into a similar pattern and there isn't a great way to manage it. Could you expand on what you had in mind for logs?

Would returning information about your `POST` request in the `GET` request be helpful? I'm imagining a `verbose` or `info` parameter on the `GET` endpoint could include the initial `filter` object in the response.

rfcs/text/0000-matrix-service.md
calvinnhieu: Ah, good catch! We do actually store both in our index and they're available to filter on, but it appears we haven't updated `GET /filters` to return this.

rfcs/text/0000-matrix-service.md
jychien: I think returning info about the POST in the GET request would definitely be helpful. Another possibility is to actually create a log file along with the downloaded matrix files. The con for that situation would be that that would be another file added to the download, and for mtx, that potentially is already a lot of files.

rfcs/text/0000-matrix-service.md
calvinnhieu: Agreed, this section could use more nuance (will update following this discussion) - the updates refer to both inconsequential and consequential metadata updates. Small clarification, updates are proposed to be responded to by the Matrix Service on a monthly cadence as opposed to being conducted by the DCP at this interval. 

The proposed 1-month period can be shortened if insufficient, but can not be real-time to preserve the cache. @lauraclarke Is there a period that would be preferred? With this being said, there are exceptions in which the Matrix Service will expedite and respond immediately to a change such as redaction. However, this should be the exception and not the rule. In the event that an update is critical, the Matrix Service is able to respond to it immediately.

Looking forward, we anticipate with the realization of data releases that both Data Accuracy and Matrix Performance can be achieved by controlling when data is synchronized across the DCP.

lauraclarke: Thanks for the clarification @calvinnhieu, I had over-interpreted your statement to suggest that updates from a wrangler/ingest perspective shouldn't be pushed on a greater frequency than once a month. 

I agree in the majority of updates a monthly cadence of responsiveness from the matrix service seems reasonable, if we end up in a position where an error gets through which might lead to significant misinterpretation of the data we can review this principle, the simplest solution, in that case, might be to disable the matrix for that project until it is fixed. 

thanks

rfcs/text/0000-matrix-service.md
lauraclarke: So I don't have access, I guess my github account has never been added to the HCA gitlab. I am concerned that a public document such as an RFC refers to private documentation. It would be wonderful if even a short summary was posted here or somewhere else publicly

@maniarathi it would be great to hear the thoughts from tech architecture about why this documentation is private and if there are good ways to expose any of it publicly

rfcs/text/0000-matrix-service.md
calvinnhieu: [Ticket](https://app.zenhub.com/workspaces/dcp-5ac7bcf9465cb172b77760d9/issues/humancellatlas/matrix-service/429). Resolving

rfcs/text/0000-matrix-service.md
calvinnhieu: [Ticket](https://app.zenhub.com/workspaces/dcp-5ac7bcf9465cb172b77760d9/issues/humancellatlas/matrix-service/430). Resolving

rfcs/text/0000-matrix-service.md
calvinnhieu: Hm yea, we have a similar issue with `loom` in which we're unable to provide additional files (e.g. summary stats, README) as users are expecting to consume the file directly without having to unzip it.

